PR #48, opened 2017-04-17
Added Git LFS GC support

DarthFennec:
A new feature that allows cleanup of Git LFS artifacts that aren't needed any longer. You pass it the path to a local Git repository on the filesystem, and it detects LFS-tracked artifacts in that repository that aren't in a remote branch, and deletes them from Artifactory. You can optionally specify exactly which refs to check to preserve artifacts, by passing either a comma-separated list of wildcard refs, or just a regular expression.


eyalbe4:
@DarthFennec,
We're currently reviewing the code. Thanks for the great work. We may modify it though before merging.

eyalbe4:
Thanks for the great contribution @DarthFennec.
We merged the functionality and it is released as part of version 1.9.0.

PR #77, opened 2017-08-04
Added support for adding priorly uploaded artifacts to local build info

rein-achten:
See https://github.com/JFrogDev/jfrog-cli-go/issues/76; added a "build-add-artifact" command that allows one to add a priorly uploaded artifact (currently one at a time) to the local build info. When one uses "build-publish" thereafter, these 'added artifacts' will be included in the build info.

What was a bit unclear to me is when to just let an error bubble up or when to use `cliutils.CheckError`. Besides, I haven't written a line of Go prior to this, so any feedback is welcome!


eyalbe4:
@rein-achten,
We'd like to better understand your use case, so that we can decide what should be the best approach to enable this functionality.
Can you please share with us how you upload the build artifacts to Artifactory? If you'd like JFrog CLI to create and publish the build-info, why not use it to also upload the artifacts? 
Thanks!

rein-achten:
@eyalbe4; we are using Ivy to publish / pull artifacts to / from an Artifactory Ivy repository. We are not using a CI tool that has a plugin for Build Integration, neither are we uploading the artifacts through the CI tool itself.
To be able to still leverage Build Integration, we thus either need to write our own scripts / tools that create and publish the build-info, or, preferably, we leverage the existing JFrog CLI. However, that currently does not allow us to add artifacts uploaded through Ivy to the build-info, which is what this pull request is about.

Hope this helps!

rein-achten:
@eyalbe4 could you give a heads-up regarding whether you're considering including such functionality or not? Thanks!

eyalbe4:
@rein-achten,
My apologies for the late response. We'll reply soon.

eyalbe4:
@rein-achten,
Thank you for sharing your use case and for your code contribution.

While discussing this feature request internally, a few concerns were raised. 
The new functionality assumes that a collection of artifacts already exists in Artifactory and attempts to tag them with a designated build name and number.

This is contrary to the existing (standard) method of publishing builds to Artifactory, which also uploads the build artifacts. The existing method minimizes the chances of errors, which lead to missing build artifacts.

Actually, the existing Checksum Deploy functionality offered by Artifactory with JFrog CLI can achieve the same goal. Checksum Deploy avoids uploading an artifact which already exists in Artifactory. It only updates their metadata. From the user's perspective, this is a regular file upload, although the file content is not transferred through the network.
In you case, even if the files are uploaded to Artifactory using a different client, "uploading" them again using JFrog CLI's upload command, can actually tag them with the desired build name and number.

I hope my response is helpful. Feel free to respond should you have any further thoughts on this matter.

rein-achten:
@eyalbe4, thanks for the update and your consideration! 

In a typical dev environment, uploading artifacts to Artifactory is way more likely to be done by other means than JFrog CLI; e.g. Ivy, Maven, docker, CI tool, etc. Being able to leverage JFrog CLI to upload build info (something that's specific to Artifactory(!)), while at the same time use the more commonly used tools to upload artifacts, I still think sounds like a logical addition :)

Regarding minimizing the chance of errors; not sure if I get the point. Ensuring that no build artifacts are missing is not something that JFrog CLI can guarantee as that information is only known to its user. It's up to the user to either (existing method) upload all artifacts while specifying both build name and build number or (new functionality) tag all existing artifacts with build name and build number. I don't see how the latter is more error prone.

The existing Checksum Deploy functionality doesn't really suit our use-case as the artifacts are only available in Artifactory at the moment we want to upload a new build info. Downloading them prior to doing a Checksum Deploy upload could work, but feels like a nasty work-around.

So, again thanks for your consideration, but it's unfortunate that you're not considering including such functionality.

eyalbe4:
Thanks @rein-achten.
Following some more brainstorming (and inspiration from your use case and suggestions) , we're considering implementing this functionality, but with a few modifications:

1. We'll add a new command named *build-add-artifacts*.
2. The command receives a wildcard pattern, as well as a build name and number.
3. The command will search for the artifacts in Artifactory, and save their details locally, to be later published, using the existing *build-publish* command.
5. Adding more artifacts and dependencies to the build is also possible using the existing *upload* and *download* commands.
6. Like most commands, the new command should also accept a File Spec.

Please let us know what you think.

rein-achten:
Thanks for the reconsideration, @eyalbe4 !

Sounds good! I'm going to try and make some time to rework the changes accordingly. Mostly, the modifications are clear. However I do have one question at this moment:

- Regarding (3); is it your intention to postpone setting of the `build name` and `build number` properties until the _build-publish_ command is being executed or did I misunderstand? That would be a nice improvement, but I guess the same thing should happen when uploading artifacts then whilst specifying `build name` and `build number`. If I understand correctly, currently the artifact is uploaded and the properties are set directly. So, if the user eventually doesn't execute the _build-publish_ command, the uploaded artifacts are tagged with a build that doesn't exit.

eyalbe4:
@rein-achten,
It is great you're willing to work on this with us!

Answering your question, it is best to tag the artifacts in Artifactory with the new build name and number only when the *build-publish* command is executed. This way, if the user changed his mind, and did not publish the build, the artifacts are kept untagged.
It is true though that the *upload* and *download* commands tag the artifacts during the upload and download. The approach of tagging them with the build info publication might be better and we might decide to make this change in the future.

Few more notes related to tagging the artifacts:
The AQL used to collect the artifacts from Artifactory, also returns the build name and number of the artifacts. We should use this, and tag only artifacts which aren't tagged or are tagged with a different build name or number. This is to avoid unnecessary REST call to Artifactory.
In addition, the new build name and number should overwrite previous name and number tags.

Please let me know if you have any other questions, thoughts or suggestions.

rein-achten:
@eyalbe4, only tagging when the build is being uploaded is indeed nicer. Will incorporate that (and if time permits extend that to the _upload_ and _download_ commands in a separate PR). Now only to find some time :)

One question though; during the implementation of the PR I created some tests that ran against an actual instance of Artifactory. In that process, I couldn't use the free version since the tests were using some features that are only available in the commercial one (I used a trial version then). We do have a corporate commercial instance running, but I don't want to pollute that with the test data. Did I miss something and is it instead possible to run the tests against the free version?

eyalbe4:
@rein-achten,

The tests indeed require an Artifactory Pro instance. This is because some of the functionality being tested relies on Artifactory Pro features. 
If you prefer not to use your Artifactory instance for the testing, you can perhaps temporarily disable some of the existing tests and leave only the tests relevant for the new feature. We'll make sure to run the entire tests before merging the code.

In addition, it is recommend to work against the *dev* branch instead of the master.

UtkarshPatkar:
Any updates on when build-add-artifacts method would be available?

rein-achten:
@UtkarshPatkar Haven't been able to free up the time to work on this PR. Hope to find some soon.

rein-achten:
Closing this one; I've created [PR-170](https://github.com/JFrogDev/jfrog-cli-go/pull/170) as a follow-up that targets the `dev` branch.

muggenhor:
> The existing Checksum Deploy functionality doesn't really suit our use-case as the artifacts are only available in Artifactory at the moment we want to upload a new build info. Downloading them prior to doing a Checksum Deploy upload could work, but feels like a nasty work-around.

@rein-achten is this really true? When producing the list of URLs (that you later add to the build with `build-add-artifact`) is it not also possible to compute the checksum at that same time?

PR #79, opened 2017-08-08
SshAuthentication: ssh-agent support

kalinchenko:
The proper condition to perform ssh authentication is that Artifactory URL starts with "ssh://".
And if --ssh-key-path is omitted, try to use ssh-agent.


eyalbe4:
Thank you for this pull request @kalinchenko.
We would like to work with you and make some changes to the code, before merging it.
Can you please join our contributors community by signing [JFrog's CLA](https://secure.echosign.com/public/hostedForm?formid=5IYKLZ2RXB543N) so that we can continue working on this together?

kalinchenko:
Ok, we sign it as company and I'll make another pull request with this patch from corporate account.

PR #80, opened 2017-08-10
SshAuthentication: ssh-agent support

kalinchenko:
The proper condition to perform ssh authentication is that Artifactory URL starts with "ssh://".
And if --ssh-key-path is omitted, try to use ssh-agent.

### DIFF
--- a/artifactory/cli.go
+++ b/artifactory/cli.go
@@ -1150,6 +1150,9 @@ func createArtifactoryDetails(c *cli.Context, includeConfig bool) (*config.Artif
 		}
 	}
 	details.Url = cliutils.AddTrailingSlashIfNeeded(details.Url)
+	if strings.HasPrefix(details.Url, "ssh://") {
+		details.IsSsh = true
+	}
 	return details, nil
 }
 
--- a/artifactory/utils/artifactoryutils.go
+++ b/artifactory/utils/artifactoryutils.go
@@ -112,7 +112,7 @@ func initTransport(artDetails *config.ArtifactoryDetails) (error) {
 }
 
 func PreCommandSetup(flags CommonFlags) (err error) {
-	if flags.GetArtifactoryDetails().SshKeyPath != "" {
+	if flags.GetArtifactoryDetails().IsSsh {
 		err = SshAuthentication(flags.GetArtifactoryDetails())
 		if err != nil {
 			return
--- a/artifactory/utils/sshagent/sshagent_unix.go
+++ b/artifactory/utils/sshagent/sshagent_unix.go
@@ -1,0 +1,17 @@
+// +build !windows
+
+package sshagent
+
+import (
+	"net"
+	"os"
+	"golang.org/x/crypto/ssh/agent"
+)
+
+func SshAgent() (agent.Agent, error) {
+	conn, err := net.Dial("unix", os.Getenv("SSH_AUTH_SOCK"))
+	if err != nil {
+		return nil, err
+	}
+	return agent.NewClient(conn), nil
+}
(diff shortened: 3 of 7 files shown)


eyalbe4:
Thanks a lot of contributing this @kalinchenko.
We'll soon start working on merging your code into the *dev* branch.

eyalbe4:
@kalinchenko, I have the following questions related to the code:
* What is the purpose of the new *escapeWinPath* function in *http.go*? I can see that it is used when constructing the download path. Where there any particular file names which required this?
* We could probably use the [ssh-agent](https://github.com/xanzy/ssh-agent) as a vendor dependency instead of copying the code in the JFrog CLI code base. Do you agree? Did you need to modify the code there?
* Did you test the new ssh-agent functionality on both Unix and Windows?

Thanks

kalinchenko:
Hi @eyalbe4, thanks for reviewing.

> What is the purpose of the new escapeWinPath function in http.go? I can see that it is used when constructing the download path. Where there any particular file names which required this?

The issue is about absolute target path in Windows platform.
In case of something like `jfrog rt dl <artifact_name> C:\some\path\package.zip` and if artifact is big enough to be downloaded by `httputils.DownloadFileConcurrently()`, the client tried to create a temporary files `<tmp_dir_path>\C:\some\path\package.zip_<chunk_number>`.
So Windows rejects creating a folder named `<tmp_dir_path>\C:` due to colon.
As for me, I don't think it is really needed to use full target path nested in temporary folder. But as simplest way to fix I preferred just to cut out colon from the path, rather than think out a proper logic of choosing temporary file names.

> We could probably use the ssh-agent as a vendor dependency instead of copying the code in the JFrog CLI code base. Do you agree? Did you need to modify the code there?

Sorry, I wasn't familiar with idea of vendor folder. Looks like my security concern (to pull dependency from untrusted source) is solved by this vendor-folder feature. On the other hand, xanzy's package is not more than copy-paste of original davidmz's package as well.
So, do you want me to use xanzy's package and commit it to vendor folder?

> Did you test the new ssh-agent functionality on both Unix and Windows?

Yes, we already use it (not widely though) on Linux and Windows platforms as custom jfrog-cli build.

eyalbe4:
Great @kalinchenko.
Thanks a lot for this!
Would you mind if we take the code in your pull request and merge it manually to dev branch? We're also going to add integration tests to test the SSH login functionality.
If however you prefer we use your commit and then add our changes on top of it please let us know.

kalinchenko:
@eyalbe4, please feel free to chose whatever fits better for you.
Thanks

eyalbe4:
@kalinchenko,
Thanks again for contributing the code and helping us add this functionality!
It is included in version 1.12.0 which has been recently released.
I'm therefore closing this issue.

kalinchenko:
@eyalbe4, thank you!
I have checked current version (1.12.1) and the problem of "absolute target path in Windows" is still there:
`[Error] open C:\<somelocalpath>\tmp\jfrog.cli.822455491\C:\<myprojpath>\<base64encodedstring>_2: The filename, directory name, or volume label syntax is incorrect.
`

Description:
https://github.com/JFrogDev/jfrog-cli-go/pull/80#issuecomment-326895076
My workaround was:
https://github.com/JFrogDev/jfrog-cli-go/pull/80/commits/78a0a82f278c31b7e0e99df8e538346a2f7eb44c

I see now you use base64-encoding of FileName to generate temporary file names, but "C:\" appears in LocalPath string, not FileName.
And base64-encoding has side-effect of increasing the string length. So in some cases it could exceed system limit.
I think it is not needed to use target path and file name to generate temporary filename at all. You can pick any random nonexistent file name for that purpose. I didn't implement this idea because didn't want to modify function interface to return that generated random name, but now you have already done that function interface change, so now it could be implemented easily.

romangurevitch:
@kalinchenko, Thank you for reporting this issue, 
We are correctly working to solve it, you can track the following [Issue](https://github.com/JFrogDev/jfrog-cli-go/issues/114) for updates.

PR #169, opened 2018-05-15
Add build-add-artifact command to support adding priorly uploaded artifacts to build info

rein-achten:
In case one uploads artifacts by other means than the JFrog CLI while
wanting to leverage JFrog CLI to upload builds, the build-add-artifact
command allows one to add priorly uploaded artifacts to the local
build info.

When publishing that build, these artifacts get tagged with the build
name and -number.

For the sake of consistency, artifacts uploaded by means of the
upload command will also only get tagged when the build is published.

Notes:
* This PR is work-in-progress. I still need to add support for spec files, add new tests and update existing tests. Before proceeding, I'd like to get some feedback on the WIP.
* This PR is the continuation of [PR-77](https://github.com/JFrogDev/jfrog-cli-go/pull/77), which I'll close in a bit.


rein-achten:
Closing, wrong target branch. Was supposed to target `dev` instead.

PR #170, opened 2018-05-15
Add build-add-artifact command to support adding priorly uploaded artifacts to build info

rein-achten:
In case one uploads artifacts by other means than the JFrog CLI while
wanting to leverage JFrog CLI to upload builds, the build-add-artifact
command allows one to add priorly uploaded artifacts to the local
build info.

When publishing that build, these artifacts get tagged with the build
name and -number.

For the sake of consistency, artifacts uploaded by means of the
upload command will also only get tagged when the build is published.

Notes:
* This PR is work-in-progress. Before proceeding, I'd like to get some feedback on the WIP. I still need to:
    * ~~Add support for spec files~~
    * Add new tests and update existing tests.
    * Prevent tagging artifacts that already tagged with the same build name and number to avoid redundant REST calls to Artifactory.
* This PR is the continuation of [PR-77](https://github.com/JFrogDev/jfrog-cli-go/pull/77), which I'll close in a bit.


jfrog-cli/artifactory/commands/npm/publish.go
```diff
@@ -161,9 +161,9 @@ func (npmp *npmPublish) doDeploy(target string, artDetails *config.ArtifactoryDe
 
 func (npmp *npmPublish) saveArtifactData() error {
 	log.Debug("Saving npm package artifact build info data.")
-	var buildArtifacts []buildinfo.Artifact
+	var buildArtifacts []buildinfo.InternalArtifact
```
rein-achten: Wasn't too sure if this would break something; this change implies that the `build.name` and `build.number` properties will be set on these artifacts at the moment the build is uploaded.

Can someone shed their light on whether or not this has nasty side-effects?

rein-achten:
Guess I'm stupid, but I can't seem to request any reviewers. Can anyone please provide some feedback?

eyalbe4:
@rein-achten,
Thank you for the pull request. We do want to see this functionality as part of the CLI.
Give us a few more days  - we'll review it soon.

rein-achten:
@eyalbe4 thanks for the quick feedback. I'm not that much of a hurry, especially given it was silent from my part for more than half a year :)

Thanks for your help, looking forward to the feedback!

jfrog-cli/artifactory/commands/buildinfo/addartifact.go
```diff
@@ -0,0 +1,70 @@
...
+	var buildArtifacts []buildinfo.InternalArtifact
+	for i := 0; i < len(addArtifactSpec.Files); i++ {
+		params, err := addArtifactSpec.Get(i).ToArtifatorySearchParams(); if err != nil {
+			return nil, err
+		}
+		resultItems, err := servicesManager.Search(&clientutils.SearchParamsImpl{ArtifactoryCommonParams: params}); if err != nil {
+			return nil, err
+		}
+		for _, resultItem := range resultItems {
+			artifactPath := resultItem.Repo + "/" + resultItem.Path + "/" + resultItem.Name
```
eyalbe4: I think there's a bug here, because *resultItem.Path* can have **.** as a value.
The *GetItemRelativePath* method of the *ResultItem* struct already takes care of it and should be probably used instead.
BTW,

rein-achten:
Thanks for the comments so far, @eyalbe4! I'll try and make some time for making changes and adding tests soon.

jfrog-cli/artifactory/cli.go
```diff
@@ -266,6 +267,18 @@ func GetCommands() []cli.Command {
 				buildDistributeCmd(c)
 			},
 		},
+		{
+			Name:      "build-add-artifact",
```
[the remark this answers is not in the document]
rein-achten: Agreed, will change.

jfrog-cli/artifactory/commands/buildinfo/addartifact.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	err = utils.SavePartialBuildInfo(flags.BuildName, flags.BuildNumber, populateFunc); if err != nil {
+		return
+	}
+
+	log.Info("Successfully added artifact to build info")
+	return
+}
+
+func getBuildArtifacts(servicesManager *artifactory.ArtifactoryServicesManager, addArtifactSpec *spec.SpecFiles) ([]buildinfo.InternalArtifact, error) {
```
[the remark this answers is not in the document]
rein-achten: Will change name to `searchArtifacts`. Will also remove `servicesManager` parameter, this is a leftover from earlier work-in-progress and its acquisition should indeed move to this function as it's not used elsewhere.

jfrog-cli/artifactory/commands/buildinfo/addartifact.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-cli/utils/config"
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-client/artifactory"
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-client/utils/errorutils"
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-client/utils/log"
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-client/artifactory/buildinfo"
+	clientutils "github.com/jfrogdev/jfrog-cli-go/jfrog-client/artifactory/services/utils"
+)
+
+func AddArtifact(addArtifactSpec *spec.SpecFiles, flags *BuildAddArtifactConfiguration) (err error) {
+	log.Info("Adding artifacts to build info " + flags.BuildName + " #" + flags.BuildNumber + "...")
```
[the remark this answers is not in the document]
rein-achten: Logging is not necessarily very consistent throughout the code; e.g. sometimes happy flow is logged whereas in other cases merely error flow is logged. I'll try and change accordingly to make things more consistent.

jfrog-cli/artifactory/commands/buildinfo/addartifact.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-cli/artifactory/spec"
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-cli/utils/config"
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-client/artifactory"
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-client/utils/errorutils"
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-client/utils/log"
+	"github.com/jfrogdev/jfrog-cli-go/jfrog-client/artifactory/buildinfo"
+	clientutils "github.com/jfrogdev/jfrog-cli-go/jfrog-client/artifactory/services/utils"
+)
+
+func AddArtifact(addArtifactSpec *spec.SpecFiles, flags *BuildAddArtifactConfiguration) (err error) {
```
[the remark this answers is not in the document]
rein-achten: Ok. However, I guess I can only either return all succeeded or all failed. Unlike with the `AddDependencies` command, I don't know the amount of artifacts referenced by the spec file until I've queried Artifactory.

jfrog-cli/artifactory/commands/buildinfo/addartifact.go
```diff
@@ -0,0 +1,70 @@
...
+	log.Info("Adding artifacts to build info " + flags.BuildName + " #" + flags.BuildNumber + "...")
+
```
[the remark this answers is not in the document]
rein-achten: Ok.

jfrog-cli/artifactory/commands/buildinfo/addartifact.go
rein-achten: Overlooked this convenience function. It seems to return the expected path, so will use this instead.

rein-achten:
@eyalbe4 Could you have another look? Before adding a test, I'd like the code to be mostly in an acceptable state.

jfrog-cli/artifactory/commands/buildinfo/publish.go
```diff
@@ -4,7 +4,10 @@ import (
 	"fmt"
 	"github.com/jfrog/jfrog-cli-go/jfrog-cli/artifactory/utils"
 	"github.com/jfrog/jfrog-cli-go/jfrog-cli/utils/config"
+        "github.com/jfrog/jfrog-cli-go/jfrog-client/artifactory"
```
rein-achten: Will fix indentation.

jfrog-cli/artifactory/commands/buildinfo/addartifacts.go
```diff
@@ -0,0 +1,73 @@
...
+		populateFunc := func(partial *buildinfo.Partial) {
+			partial.Artifacts = buildArtifacts
+		}
+		err = utils.SavePartialBuildInfo(configuration.BuildName, configuration.BuildNumber, populateFunc); if err != nil {
+			return 0, len(buildArtifacts), err
+		}
+
+		return len(buildArtifacts), 0, err
+	} else {
+		return 0, 0, nil
```
eyalbe4: There's no reason to always return 0, 0 if dry-run is used. The information about the total of affected files is useful also  with dry-run. You look at the *AddDependencies* function (used by the build-add-dependencies command) as a reference. It makes sense for the two commands (build-add-dependencies and build-add-artifacts) to be consistent in terms of behavior.

jfrog-cli/artifactory/commands/buildinfo/addartifacts.go
```diff
@@ -0,0 +1,73 @@
...
+		for _, resultItem := range resultItems {
+			artifactPath := resultItem.GetItemRelativePath()
+			artifact := buildinfo.InternalArtifact{Path: artifactPath, Checksum: &buildinfo.Checksum{}}
+			artifact.Sha1 = resultItem.Actual_Sha1
+			artifact.Md5 = resultItem.Actual_Md5
+			buildArtifacts = append(buildArtifacts, artifact)
+		}
+	}
+
+	if len(buildArtifacts) == 0 {
```
eyalbe4: We should not automatically throw an error here. An error should be thrown if the *fail-no-op* option is set to true. You can use the code in *cli.go* to take care of this for you.
The fact the option does not exist for the *build-add-dependencies* is a bug. We'll soon add it.

jfrog-cli/artifactory/commands/buildinfo/publish.go
```diff
@@ -225,6 +235,22 @@ func createExcludeFilter(pattern string) filterFunc {
...
+	for _, partial := range partials {
+		switch {
+		case partial.Artifacts != nil:
+			for _, artifact := range partial.Artifacts {
+				resultItems = append(resultItems, rtclientutils.ResultItem{Path: artifact.Path})
+			}
+		}
+	}
+
+	props := "build.name=" + buildInfo.Name + ";build.number=" + buildInfo.Number
```
eyalbe4: We should not build the build properties string by ourselves, but use the existing *CreateBuildProperties* API to do it for us. It also tales care of adding the *build.timestamp* property to the string.

jfrog-cli/artifactory/commands/buildinfo/publish.go
```diff
@@ -225,6 +235,22 @@ func createExcludeFilter(pattern string) filterFunc {
 	}
 }
 
+func setBuildInfoPropertiesForArtifacts(servicesManager *artifactory.ArtifactoryServicesManager, buildInfo *buildinfo.BuildInfo, partials buildinfo.Partials) error {
+	var resultItems []rtclientutils.ResultItem
+	for _, partial := range partials {
```
eyalbe4: This switch case can be simplified to:
*if partial.Artifacts != nil {*

jfrog-cli/artifactory/commands/buildinfo/publish.go
```diff
@@ -17,7 +20,12 @@ func Publish(buildName, buildNumber string, config *buildinfo.Configuration, art
 		return err
 	}
 
-	buildInfo, err := createBuildInfoFromPartials(buildName, buildNumber, config, artDetails)
+	buildInfo, partials, err := createBuildInfoFromPartials(buildName, buildNumber, config, artDetails)
+	if err != nil {
+		return err
+	}
+
```
eyalbe4: Why are doing *setBuildInfoPropertiesForArtifacts* are part fo the *build-publish* command?
This actually adds a setProps request to Artifactory for every published build (even if the files had already been tagged). The *build-add-artifacts* command should do tag the artifacts (set the props) by itself.

eyalbe4:
@rein-achten,
Will you be available to continue working on this?

rein-achten:
@eyalbe4 I have a hard time finding the time at this moment. Will try to continue working on this soon.

jfrog-cli/artifactory/commands/buildinfo/addartifacts.go
rein-achten: In a dry-run, I shouldn't actually query the Artifactory server, right? Since querying Artifactory is necessary to get a list of affected artifacts (as opposed to the `build-add-dependencies` command that obtains this from the local file system), I don't see how I can return anything else.

eyalbe4:
@rein-achten,
Would you like me review the code again? If so, feel free to resolve the conflicts and let me know.
I'll try to be responsive so that we can push this forward faster.

eyalbe4:
@rein-achten,
I'm closing this PR for now. We can always recreate the PR and resume this the work on this initiative.

PR #174, opened 2018-05-29
support artifactory ping

galusben:


jfrog-cli/jfrog/artifactory_test.go
```diff
@@ -2108,6 +2108,11 @@ func TestGitLfsCleanup(t *testing.T) {
 	cleanArtifactoryTest()
 }
 
+func TestPing(t *testing.T) {
+	initArtifactoryTest(t)
+	artifactoryCli.Exec("ping")
```
yahavi: I think you forgot
cleanArtifactoryTest()

yahavi: One more thing: Can you add also a negative test? i.e. test against a non-exists Artifactory.

galusben: Look at https://github.com/JFrogDev/jfrog-cli-go/pull/174/files#diff-446bb60625c66b8d69abdf8a3087016a

PR #184, opened 2018-07-02
Ability to specify version with default set to latest

ankushchadha:
In dockerfile, it is recommended to specify a version of tools that need to be installed instead of using latest. 

Updated the script to allow an option to specify a version. 

Usage: cat getCli.sh | sh -s -- 1.16.0 will download 1.16.0 version vs cat getCli.sh | sh that downloads latest. Tested on mac and ubuntu

### DIFF
--- a/getCli.sh
+++ b/getCli.sh
@@ -3,23 +3,31 @@
 CLI_OS="na"
 CLI_UNAME="na"
 
+if [ $# -eq 0 ]
+  then
+	VERSION=\$latest
+	echo "Using latest version of jfrog cli"
+  else
+	VERSION=$1
+fi
+
 if $(echo "${OSTYPE}" | grep -q msys); then
     CLI_OS="windows"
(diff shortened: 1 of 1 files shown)


getCli.sh
```diff
@@ -3,26 +3,34 @@
 CLI_OS="na"
 CLI_UNAME="na"
 
+if [ $# -eq 0 ]
+  then
+	VERSION=\$latest
+	echo "Using latest version of jfrog cli"
```
eyalbe4: I suggest slightly modifying the message to:
**echo "Downloading the latest version of JFrog CLI..."**

getCli.sh
```diff
@@ -3,26 +3,34 @@
...
+  else
+	VERSION=$1
```
eyalbe4: To be consistent, let's add another echo here prompting:
**echo "Downloading version $1 of JFrog CLI..."**

ankushchadha:
Thanks, @eyalbe4 
This will help to download a specific version of JFrog CLI in order to build a statically tagged version of a builder image that can be used to build applications via JFrog CLI.

PR #212, opened 2018-09-02
Fix lock unit test

yahavi:


jfrog-cli/utils/lock/lock_test.go
```diff
@@ -3,15 +3,16 @@ package lock
...
 	"fmt"
 	"github.com/jfrog/jfrog-cli-go/jfrog-client/utils/io/fileutils"
+	"math"
 	"os"
 	"testing"
 	"time"
 )
 
-func TestLock(t *testing.T) {
+func TestLockPidSmaller(t *testing.T) {
```
eyalbe4: I suggest we add a description comment to this function, mainly descriving the difference between this function and *TestLockPidBigger*.
I think we should add a similar description to the *TestLockPidBigger* function. In both description we can briefly explain why the PID matters and when.
I suggest renaming *TestLockPidSmaller* to *TestLockSmallerP{id* and *TestLockPidBigger* to *TestLockLargerPid*.

eyalbe4:
Awesome @yahavi!

PR #213, opened 2018-09-03
Deprecate no registry flag and mask CMD output

AlexeiVainshtein:


jfrog-cli/artifactory/commands/golang/go.go
```diff
@@ -93,3 +104,24 @@ func validatePrerequisites() error {
 	}
 	return nil
 }
+
+/*
+Resolves the dependencies from the official Go repositories and publish those dependencies to Artifactory
```
eyalbe4: Let's use // instead of /*...*/ for comments, to be consistent with the existing code.
There are a few other such comments in this PR which I suggest fixing.

The method should named as a private method, because it is not used outside this file. I renaming it to *downloadAndPublish*.

I suggest rephrasing the function comment to:
*Download the dependencies from VCS and publish them to Artifactory.*

jfrog-cli/artifactory/commands/golang/go.go
```diff
@@ -71,7 +73,16 @@ func ExecuteGo(noRegistry bool, goArg, targetRepo, buildName, buildNumber string
 	}
 	err := goutils.RunGo(goArg)
 	if err != nil {
-		return err
+		if !noRegistry && strings.EqualFold(err.Error(), "404 Not Found") {
+			// Need to run Go without Artifactory to resolve all dependencies.
+			log.Debug("Got", err.Error(), "from", details.GetUrl()+"api/go/"+targetRepo+".", "Trying resolving directly and publishing the dependencies to Artifactory")
```
eyalbe4: I suggest changing this log message from Debug to Info. I also suggest rephrasing it a bit to be:
*log.Info("Received", err.Error(), "from Artifactory.", "Trying download the dependencies for the VCS and upload them to Artifactory...")*

jfrog-cli/artifactory/commands/golang/go.go
```diff
@@ -93,3 +104,24 @@ func validatePrerequisites() error {
...
+func TryToDownloadAndPublish(targetRepo string, details *config.ArtifactoryDetails) error {
+	err := os.Unsetenv(goutils.GOPROXY)
+	if err != nil {
+		return errorutils.CheckError(err)
+	}
+	err = goutils.DownloadDependenciesDirectly()
+	if err != nil {
+		return err
+	}
+	// Publish the dependencies.
```
eyalbe4: Let's add a log message here:
*log.Info("Uploading dependencies to Artifactory...")*

jfrog-cli/artifactory/utils/golang/cmd.go
```diff
@@ -79,5 +81,37 @@ func RunGo(goArg string) error {
 	if err != nil {
 		return errorutils.CheckError(err)
 	}
+
+	regExp, err := utils.GetRegExp("((http|https)://)")
```
eyalbe4: For performance reasons, let's compile the regexp only once. This improvement will be relevant in cases when we run multiple go commands.

jfrog-cli/artifactory/utils/utils.go
```diff
@@ -158,9 +161,111 @@ func RunCmd(config CmdConfig) error {
...
+	}
+
+	return excludePathsRegExp, nil
+}
+
+/*
+Mask the credentials information from the line. The credentials are build as user:password
+For example: http://user:password@127.0.0.1:8081/artifactory/path/to/repo
+*/
+func MaskCredentials(line string) (string, error) {
```
eyalbe4: Let's see fi we can come up with a more generic implementation, which can do the masking according to a regular expression.

jfrog-cli/artifactory/utils/golang/cmd.go
AlexeiVainshtein: It's being compiled only once. When running Go Cmd.

jfrog-cli/artifactory/commands/golang/go.go
AlexeiVainshtein: This message already printed:

[Info] Publishing package dependencies...
[Info] Publishing: gopkg.in/go-playground/validator.v8:v8.18.2 to go-local2

This message from the publish function:
	// Publish the dependencies.
	_, _, err = Publish(false, "ALL", targetRepo, "", "", "", details) --> PublishDependencies:
	log.Info("Publishing package dependencies...")

jfrog-cli/artifactory/cli.go
```diff
@@ -1317,8 +1317,11 @@ func goCmd(c *cli.Context) {
 	details := createArtifactoryDetailsByFlags(c, true)
 
 	logGoVersion()
-
-	err := golang.ExecuteGo(c.Bool("no-registry"), goArg, targetRepo, buildName, buildNumber, details)
+	noRegistry := c.Bool("no-registry")
+	if noRegistry {
+		log.Warn("no-registry flag is deprecated and will be removed in upcoming releases. Use the command without the flag instead.")
+	}
+	err := golang.ExecuteGo(noRegistry, goArg, targetRepo, buildName, buildNumber, details)
```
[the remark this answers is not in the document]
AlexeiVainshtein: Done

jfrog-cli/artifactory/cli.go
```diff
@@ -711,7 +711,7 @@ func getGoFlags() []cli.Flag {
 	flags := []cli.Flag{
 		cli.BoolFlag{
 			Name:  "no-registry",
-			Usage: "[Default: false] Set to true if you don't want to use Artifactory as your proxy",
+			Usage: "[Default: false] Deprecated. Use the command without this flag. Set to true if you don't want to use Artifactory as your proxy",
```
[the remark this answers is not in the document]
AlexeiVainshtein: Done

jfrog-cli/artifactory/commands/golang/go.go
AlexeiVainshtein: Done

jfrog-cli/artifactory/commands/golang/go.go
AlexeiVainshtein: Done

jfrog-cli/artifactory/utils/utils.go
```diff
@@ -215,57 +215,31 @@ type CmdConfig interface {
...
-	}
-	if strings.HasSuffix(line, "//") {
-		line = line[:len(line)-2]
-	}
-
-	return line, nil
+// Mask the credentials information from the line. The credentials are build as user:password
+// For example: http://user:password@127.0.0.1:8081/artifactory/path/to/repo
+func (reg *RegExpStruct) MaskCredentials() (string, error) {
+	splittedResult := strings.Split(reg.matcher, reg.Separator)
```
eyalbe4: I'm not sure we need the split here. LMK what you think.

jfrog-cli/artifactory/utils/utils_test.go
```diff
@@ -23,24 +23,64 @@ func TestGetHomeDir(t *testing.T) {
 }
 
 func TestRemoveCredentialsFromURL(t *testing.T) {
```
eyalbe4: This comment applies to both the *TestRemoveCredentialsFromURL* and * TestReturnErrorOnNotFound* functions:
In case one of the tests fails, I'm not sure it will be easy to understand from the test error message, which line pattern failed. What do you think?

AlexeiVainshtein: I've changed the tests to also provide the regex result and also the output of the line after masking.

PR #215, opened 2018-09-03
Support retries in upload command

DimaNevelev:


eyalbe4:
I've gone through the code. Looks good @DimaNevelev.

BrBill:
As the _(a?)_ requester for this, you have my thanks for this, looking forward to trying the fix.

scottrus:
Will this retry also apply to downloads or is it strictly uploads? We EOF errors during simple network events that will stop a download. If downloads can be retried that would be be a great solution.

eyalbe4:
@scottrus,
Download retries ae already supported. The *jfrog rt download* command has the *--retry* option with a default value of 3. You can set it to a higher value if needed.

scottrus:
Thanks @eyalbe4. That seems a bit odd that we see an EOF even after 3 retries then. I'll look into it more and file an issue if I think it's needed.

eyalbe4:
Sounds good @scottrus. Your feedback can be helpful. See that the JFrog CLI version you're using includes the *--retry* option.

PR #216, opened 2018-09-05
Tests in appveyor

yahavi:


jfrog-cli/utils/tests/utils.go
```diff
@@ -37,6 +37,8 @@ var TestGo *bool
 var DockerRepoDomain *string
 var DockerTargetRepo *string
 var TestNuget *bool
+var MaskCredentials *bool
```
eyalbe4: I think that the *MaskCredentials* option is redundant. Let's always mask the credentials.

jfrog-cli/utils/tests/utils.go
```diff
@@ -215,17 +219,24 @@ func NewJfrogCli(mainFunc func(), prefix, suffix string) *JfrogCli {
...
 			continue
 		}
-		os.Args = append(os.Args, strings.Split(v, spaceSplit)...)
+		args := strings.Split(v, spaceSplit)
+		os.Args = append(os.Args, args...)
+		output = append(output, args...)
 	}
 	if cli.suffix != "" {
-		os.Args = append(os.Args, strings.Split(cli.suffix, spaceSplit)...)
+		args := strings.Split(cli.suffix, spaceSplit)
```
eyalbe4: cre suggest renaming *cli.suffix* to *cli.credentials* (because it only holds the credentials).
Let's also rename *cli.prefix* to *cli.args*.

PR #227, opened 2018-09-25
Support for Artifactory Go repositories new publish API

AlexeiVainshtein:


jfrog-client/artifactory/services/go/go.go
```diff
@@ -51,12 +63,37 @@ func addPropertiesHeaders(props string, headers *map[string]string) error {
...
+	if strings.HasSuffix(*url, ";") {
+		tempUrl := *url
+		tempUrl = tempUrl[:len(tempUrl)-1]
+		*url = tempUrl
+	}
+	return nil
+}
+
+// Returns true if needed to use properties as header or false if need to use matrix params.
+func shouldUseHeaders(artifactoryVersion string) bool {
```
eyalbe4: I suggest adding a unit test for this function.
Also, to make the code a bit more easier to read, I suggest adding to the above function description, the min Artifactory version for which this function returns *true*.

eyalbe4:
Looks good @AlexeiVainshtein!
Added one inline suggestion.

jfrog-client/artifactory/services/go/go.go
AlexeiVainshtein: Done. I've added a test a changed the documentation.

Issue #229, opened 2018-10-07
Add the ability to upload broken symlinks.


yudkeh:
Sometimes there are symlink which are broken in the machine which used for upload but when they will get downloaded by the users they will work as they include path to an existing folder, therefore there is a need to add the ability to upload broken symlink.

trinhta:
Currently the broken symlinks are totally ignored during the upload, but the ideal solution for broken symlinks is that Artifactory should create a "marker" in Artifactory for the bad symlinks during the upload so that later when the user downloads from Artifactory, using the --symlink option, the same (bad) symlinks will appear in his workspace.

RobiNino:
Hi @trinhta @yudkeh, 
We are currently working on a solution to this issue, and will release it as soon as possible.

RobiNino:
Hi @trinhta,
CLI version 1.21.0 is now released and includes a fix for this issue.
We would love to hear your feedback regarding this matter.
@yudkeh

PR #237, opened 2018-10-17
Upload broken symlink fixed

RobiNino:
Fixed jfrog-cli-go #229 

Added the ability to upload broken symlinks with --symlinks=true.

Fixed an issue where when --symlinks=false, the symlink target would be uploaded given the symlink name instead of the target name.

This pull request has a matching pull request in jfrog-client-go.

### DIFF
--- a/jfrog-cli/jfrog/artifactory_test.go
+++ b/jfrog-cli/jfrog/artifactory_test.go
@@ -944,6 +944,27 @@ func TestExcludeBrokenSymlink(t *testing.T) {
 	cleanArtifactoryTest()
 }
 
+// Testing uploading broken symlinks.
+// Uploaded symlinks are not verified by default, so all symlinks should be uploaded
+func TestUploadBrokenSymlink(t *testing.T) {
+	if cliutils.IsWindows() {
+		return
+	}
+	initArtifactoryTest(t)
+
+	// Creating broken symlink
+	os.Mkdir(tests.Out, 0777)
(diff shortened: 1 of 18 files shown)


jfrog-cli/jfrog/artifactory_test.go
```diff
@@ -944,6 +944,27 @@ func TestExcludeBrokenSymlink(t *testing.T) {
 	cleanArtifactoryTest()
 }
 
+// Testing uploading broken symlinks.
+// Uploaded symlinks are not verified by default, so all symlinks should be uploaded
```
eyalbe4: I think that **by default** should be removed from the comment, because symlinks are never verified by default.

jfrog-cli/jfrog/artifactory_test.go
```diff
@@ -944,6 +944,27 @@ func TestExcludeBrokenSymlink(t *testing.T) {
...
+func TestUploadBrokenSymlink(t *testing.T) {
```
eyalbe4: I suggest modifying this test from a "TestUploadBrokenSymlink" to a "TestValidateSymlink".
Basically, the test can upload a broken symlink (the same way it does now) and then, verify that this symlink fails to be downloaded with the *--validate-symlink* option and succeeds to download it without the option.
This way, instead of testing a bug fix, we're testing a real functionality of JFrog CLI.

Issue #172, opened 2018-05-24
JFROG_CLI_HOME misinterpreted


kPshi:
While working with different config files via JFROG_CLI_HOME I found that another /.jfrog is appended to the configured directory.
```
jfrog --help
[...]
	JFROG_CLI_HOME
		[Default: ~/.jfrog]
		Defines the JFrog CLI home directory path.
```
so actually setting it to the default value should not change anything, but
```
export JFROG_CLI_HOME=~/.jfrog

strace 2>&1 jfrog rt u --recursive swshare{,} | grep /\\.jfrog
stat64("/home/x/.jfrog/.jfrog", 0x18e22384) = -1 ENOENT (No such file or directory)
stat64("/home/x/.jfrog", {st_mode=S_IFDIR|0775, st_size=4096, ...}) = 0
mkdirat(AT_FDCWD, "/home/x/.jfrog/.jfrog", 0777) = 0
stat64("/home/x/.jfrog/.jfrog/jfrog-cli.conf", 0x18e224a4) = -1
```
So that either the documentation (--help) is incorrect or the variable is not interpreted correctly. Personally I think it's the latter.

Using jfrog CLI 1.16.0 on Ubuntu 16.04.4 LTS

eyalbe4:
Than you for letting us know @kPshi.
We consider this a bug and we intend to fix it. We'll fix it in a way that will not affect anyone who's already using this environment variable. 
We'll update this issue after fixing the issue.

RobiNino:
Hi @kPshi,
A new environment variable, JFROG_CLI_HOME_DIR, was introduced in Jfrog CLI version 1.21.0.
This variable is interpreted correctly and replaces the old JFROG_CLI_HOME variable.

Thanks for pointing this issue out, we'll appreciate your feedback regarding this matter.

PR #238, opened 2018-10-17
New environment variable, JFROG_CLI_HOME_DIR, to replace JFROG_CLI_HOME

RobiNino:
Introduced a new environment variable to replace JFROG_CLI_HOME that was misinterpreted (#172).
Only the new environment variable should be used, however the old one is still supported (so that the cli won't break for users that rely on the old var).

### DIFF
--- a/jfrog-cli/utils/config/config.go
+++ b/jfrog-cli/utils/config/config.go
@@ -22,12 +22,12 @@ import (
 
 // This is the default server id. It is used when adding a server config without providing a server ID
 const (
-	DefaultServerId = "Default-Server"
-	JfrogHomeDirEnv = "JFROG_CLI_HOME_DIR"
-	JfrogHomeEnv    = "JFROG_CLI_HOME" // Old environment variable.
-	// Only used to support users relying on it before introducing JfrogHomeDirEnv
+	DefaultServerId   = "Default-Server"
+	JfrogHomeDirEnv   = "JFROG_CLI_HOME_DIR"
 	JfrogConfigFile   = "jfrog-cli.conf"
 	JfrogDependencies = "dependencies"
+	// Deprecated:
+	JfrogHomeEnv = "JFROG_CLI_HOME"
 )
 
 func IsArtifactoryConfExists() (bool, error) {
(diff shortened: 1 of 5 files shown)


jfrog-cli/utils/config/config.go
```diff
@@ -17,13 +18,14 @@ import (
...
 )
 
 // This is the default server id. It is used when adding a server config without providing a server ID
 const (
-	DefaultServerId   = "Default-Server"
-	JfrogHomeEnv      = "JFROG_CLI_HOME"
+	DefaultServerId = "Default-Server"
+	JfrogHomeDirEnv = "JFROG_CLI_HOME_DIR"
+	JfrogHomeEnv    = "JFROG_CLI_HOME" // Old environment variable.
+	// Only used to support users relying on it before introducing JfrogHomeDirEnv
```
eyalbe4: It is best to put the description comment above the property it described. As it is now, it is a bit confusing, and may look like it is actually describing the *JfrogConfigFile* property.
In addition, in this specific case, it may be enough to have only
// Deprecated
as the description.

jfrog-cli/utils/config/config.go
```diff
@@ -239,7 +241,14 @@ func convertIfNecessary(content []byte) ([]byte, error) {
 }
 
 func GetJfrogHomeDir() (string, error) {
-	if os.Getenv(JfrogHomeEnv) != "" {
+	// If JfrogHomeDirEnv exists, use it.
+	// Else if JfrogHomeEnv exists, use it by adding .jfrog at the end.
+	// The second variable was left only to support users relying on it before introducing JfrogHomeDirEnv,
+	// shouldn't be used otherwise.
+
```
eyalbe4: I suggest simplifying the above description comment and replace it with:
// The JfrogHomeEnv environment variable has been deprecated and replaced with JfrogHomeDirEnv 

I also suggest renaming the *userDir* local variable in this function to *userHomeDie*.

eyalbe4:
Looks good @RobiNino!
See my inline suggestions.

PR #255, opened 2018-11-07
Build info wrong artifact name - bug fix

RobiNino:
Fixed the following issues:

- Artifact's local path was constructed incorrectly.
- The artifact's file name for build info was taken from local path instead of artifactory path.

This pull requests has a complementing one in jfrog-client-go


eyalbe4:
@RobiNino  - 👍

eyalbe4:
Great catch @RobiNino!
While looking into this I noticed the following:
The *fileinfoutils.go* and *utils/utils.go* (which include the *StripThreadId* function) shouldn't be part of the *jfrog-client-go* library, but should be actually included in *jfrog-cli-go*. The reason of this is that these utils are needed by JFrog CLI only and are not actually related to the client functionality.

PR #268, opened 2018-11-11
Panic when --spec-vars ends with semicolon - Bug fix

### LINKED ISSUE #253: Panic when --spec-vars ends with semicolon
CLI goes into panic when `--spec-vars` option ends with semicolon(`;`). I'm using version `1.20.1`

Command giving the panic is:
`jfrog rt u --spec /path/to/upload/spec.json --spec-vars "foo=bar;"`

Error: 

`panic: runtime error: index out of range`

It seems issue is with this [line](https://github.com/jfrog/jfrog-cli-go/blob/1.20.1/jfrog-cli/utils/cliutils/utils.go#L172) .

RobiNino:
Fixes #253.


eyalbe4:
@RobiNino 👍

Issue #83, opened 2017-08-23
Support SSH-Keys with passphrase


Rondom:
I just tried to authenticate with Artifactory using my SSH-key and I got the following message, because my key has a passphrase.

    [Info] Performing SSH authentication...
    [Error] ssh: cannot decode encrypted private keys

I think it is not uncommon to protect SSH-Keys using a passphrase.

Expected: jfrog-cli should ask for a passphrase and ideally use the system ssh-agent to support more advanced scenarios than a simple passphrase as well. For a start asking for the passphrase instead of printing that message, would already help.

jgeorgeson:
Please support encrypted keys!

https://stackoverflow.com/questions/42105432/how-to-use-an-encrypted-private-key-with-golang-ssh

jgeorgeson:
Looks like this is fixed in 1.12! Thanks!

eyalbe4:
1.12 does fixes this.
@Rondom, we'd also appreciate your feedback after you try this.

Rondom:
I will try on Monday.

One thing I notice from the commits is that you accept a password from the command line. Many tools that support supplying passwords non-interactively, read them from stdin instead of from command line for security reasons. Especially on shared machines, supplying secrets via command line arguments is an issue, because everyone can see the cmdline including password by running `ps aux` while the tool is running!

eyalbe4:
@Rondom,
If you don't add the passphrase as command option, and your RSA key is encrypted, the CLI will interactively prompt you to enter the passphrase. If you're using a SSH agent, the CLI will obtain the passphrase from it.

Rondom:
The handling of encrypted keys works perfectly. Thank you!

The only thing I noticed: When you run `jfrog rt config` it will ask you whether you want to use SSH agent or not. So you need to know in advance whether you want to use an SSH agent or not. If no agent is available, you get `[Error] SSH agent requested but SSH_AUTH_SOCK not-specified`. But this is not always a given that the agent is always available. For example, in the default configuration of many Linux distribution an SSH-agent will be started automatically for GUI-sessions, but not for text-only sessions such as when logging in via SSH. SSH uses an agent if it is availabe, i.e. it falls back to a password-prompt and only uses the agent if it is available.
What is the rationale for having this configurable? I guess it will also fail, when one wants to use SSH-Agent for a Key that is in non-standard location, but I have not tried.

yahavi:
@Rondom,
Thanks for your feedback.
[...]

PR #270, opened 2018-11-12
SSH Authentication behaviour update

RobiNino:
Following Issue #83, the behaviour of ssh authentication was changed to:

Try authenticating via ssh-agent.
If failed due to any reason (No agent / connection failed / no key in agent):
Try authenticating via ssh-key.
If the key is encrypted, passphrase has to be passed via the --ssh-passphrase flag.


jfrog-cli/artifactory/commands/config.go
```diff
@@ -145,20 +137,44 @@ func getConfigurationFromUser(details, defaultDetails *config.ArtifactoryDetails
...
+func getSshKeyPath(details *config.ArtifactoryDetails) error {
+	// If path not provided as a key, read from console:
 	if details.SshKeyPath == "" {
-		ioutils.ScanFromConsole("SSH key file path", &details.SshKeyPath, savedDetails.SshKeyPath)
+		ioutils.ScanFromConsole("SSH key file path (optional)", &details.SshKeyPath, "")
+	}
+
+	// If path still not provided, return and warn about relying on agent.
+	if details.SshKeyPath == "" {
+		log.Info("SSH Key path not provided. Specify a key path using flags, or rely on ssh-agent only.")
```
eyalbe4: I suggest rephrasing the message to:
SSH Key path not provided. You can also specify a key path using the --ssh-ket-path command option.

jfrog-cli/artifactory/commands/config.go
```diff
@@ -145,20 +137,44 @@ func getConfigurationFromUser(details, defaultDetails *config.ArtifactoryDetails
...
+	if exists {
+		sshKeyBytes, err := ioutil.ReadFile(details.SshKeyPath)
+		if err != nil {
+			return nil
+		}
+		encryptedKey, err := auth.IsEncrypted(sshKeyBytes)
+		// If exists and not encrypted (or error occurred), return without asking for passphrase
+		if err != nil || !encryptedKey {
+			return err
+		} else {
```
eyalbe4: Else after return is redundant.

jfrog-cli/artifactory/commands/config.go
```diff
@@ -145,20 +137,44 @@ func getConfigurationFromUser(details, defaultDetails *config.ArtifactoryDetails
...
+			log.Info("The key file at the specified path is encrypted.")
+			log.Info("Please pass the ssh-passphrase as a flag when authentication required.")
+		}
+	} else {
+		log.Info("Could not find key in provided path. You may place the key file there later.")
+		log.Info("If you choose to use an encrypted key, you may pass the passphrase as a flag when authentication required.")
```
eyalbe4: 1. We can merge the above to log info messages into one message,
2. I suggest rephrasing:
"...you may pass the passphrase as a flag when authentication required."
to:
"...you may pass the passphrase as an option with every command."

The above two comments also applies to the other 2 log messages in this function.

jfrog-cli/artifactory/cli.go
```diff
@@ -1060,7 +1060,12 @@ func getConfigFlags() []cli.Flag {
...
 		},
 	}
-	return append(flags, getCommonFlags()...)
+	flags = append(flags, getBaseFlags()...)
+	return append(flags,
+		cli.StringFlag{
+			Name:  "ssh-key-path",
+			Usage: "[Optional] SSH key file path.",
+		})
 }
```
eyalbe4: Since the ssh-key-path flag is defined twice in this file, let's move its creation into a function and reuse it in these two places.

eyalbe4:
Nice work @RobiNino!
See my inline comments.

PR #280, opened 2018-11-22
Remove redundant 'value' word from commands help

RobiNino:
Using placeholders to remove the 'value' word from appearing on help


eyalbe4:
👍 @RobiNino.

PR #287, opened 2018-12-09
Add 'update-bundle' and 'update-bundle-config' commands

yahavi:


yahavi:
https://github.com/jfrog/jfrog-client-go/pull/19

PR #300, opened 2018-12-20
Change regex to match quotation marks in mod file

AlexeiVainshtein:


jfrog-cli/artifactory/utils/golang/project/project.go
```diff
@@ -346,7 +346,7 @@ func (project *goProject) archiveProject(version string) (string, error) {
 
 // Parse module name from go.mod content.
 func parseModuleName(modContent string) (string, error) {
-	r, err := regexp.Compile(`module ([\w\.@:%_\+-.~#?&]+/?.+)`)
+	r, err := regexp.Compile(`module "?([\w\.@:%_\+-.~#?&]+/?.+\w)`)
```
eyalbe4: This reg exp is being compiled multiple times throughout the execution flow. The compilation of a reg exp is an expensive operation.
We should probably compile it only once, do it lazily and put it in a util, because it is the same for all projects.

eyalbe4:
👍 @AlexeiVainshtein

PR #305, opened 2019-01-06
New feature: version command shows if newer version are available

retgits:
JFrog CLI is awesome, but it doesn't show if there are any new versions available. This PR introduces that behavior:

```bash
$ jfrog-cli
NAME:
   jfrog - See https://github.com/jfrog/jfrog-cli-go for usage instructions.

USAGE:
   main [global options] command [command options] [arguments...]

VERSION:
   1.23.1 (you're running the latest version)

... snip ...
```

When you're running the latest version:

```bash
$ jfrog-cli --version
jfrog version 1.23.1 (you're running the latest version)
```

When there is a newer version available

```bash
$ jfrog-cli --version
jfrog version 1.0.0 (there is a newer version available [v1.23.1])
```

As not everyone is interested in these messages there is also a new environment variable `JFROG_CLI_SHOW_UPDATE`

```
JFROG_CLI_SHOW_UPDATE
[Default: true]
If true, JFrog CLI shows whether you're running the latest version.
```

If this variable is set to `false`, the behavior of the CLI is exactly as it is currently. If this variable is not set, or set to `true`, it will display the new message as above

### DIFF
--- a/jfrog-cli/docs/common/help.go
+++ b/jfrog-cli/docs/common/help.go
@@ -16,4 +16,8 @@ const GlobalEnvVars string = `	JFROG_CLI_LOG_LEVEL
 	JFROG_CLI_HOME_DIR
 		[Default: ~/.jfrog]
 		Defines the JFrog CLI home directory path.
+	
+	JFROG_CLI_SHOW_UPDATE
+		[Default: true]
+		If true, JFrog CLI shows whether you're running the latest version.
 		`
--- a/jfrog-cli/docs/common/help.go
+++ b/jfrog-cli/docs/common/help.go
@@ -17,7 +17,7 @@ const GlobalEnvVars string = `	JFROG_CLI_LOG_LEVEL
 		[Default: ~/.jfrog]
 		Defines the JFrog CLI home directory path.
 	
-	JFROG_CLI_SHOW_UPDATE
+	JFROG_CLI_SHOW_VERSION_UPDATE
 		[Default: true]
 		If true, JFrog CLI shows whether you're running the latest version.
 		`
--- a/jfrog-cli/utils/cliutils/utils.go
+++ b/jfrog-cli/utils/cliutils/utils.go
@@ -1,15 +1,19 @@
 package cliutils
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"runtime"
+	"strings"
+
 	"github.com/codegangsta/cli"
 	"github.com/jfrog/jfrog-cli-go/jfrog-cli/utils/summary"
 	"github.com/jfrog/jfrog-client-go/utils"
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 	"github.com/jfrog/jfrog-client-go/utils/log"
-	"os"
-	"runtime"
-	"strings"
 )
 
 // Error modes (how should the application behave when the CheckError function is invoked):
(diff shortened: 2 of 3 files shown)


jfrog-cli/utils/cliutils/utils.go
```diff
@@ -114,9 +118,51 @@ func confirmAnswer(answer string) bool {
...
+
+		if CliVersion == latestRelease {
+			return fmt.Sprintf("%s (you're running the latest version)", CliVersion)
+		}
+		return fmt.Sprintf("%s (there is a newer version available [v%s])", CliVersion, latestRelease)
+	}
 	return CliVersion
 }
 
+func getLatestRelease() (string, error) {
```
eyalbe4: 1. It is probably best to get the latest JFrog CLI version from Bintray,
2. We prefer to use JFrog CLI's http client infrastructure, rather than duplicating the functionality. Since we'll be using Bintray to get the latest version, then we can add a new GetLatestVersion API to the jfrog-client-go library used by jfrog-cli-go. This new API can be then used by anyone, similarly to all the other APIs documented in the [jfrog-client-go's README](https://github.com/jfrog/jfrog-client-go/blob/master/README.md).

jfrog-cli/docs/common/help.go
```diff
@@ -16,4 +16,8 @@ const GlobalEnvVars string = `	JFROG_CLI_LOG_LEVEL
 	JFROG_CLI_HOME_DIR
 		[Default: ~/.jfrog]
 		Defines the JFrog CLI home directory path.
+	
+	JFROG_CLI_SHOW_UPDATE
```
eyalbe4: I suggest renaming it to JFROG_CLI_SHOW_VERSION_UPDATE.

eyalbe4:
Thank you for this @retgits!
See my inline suggestions.

jfrog-cli/docs/common/help.go
retgits: Thanks for the feedback, totally agree with changing the name.

jfrog-cli/utils/cliutils/utils.go
retgits: Great feedback, absolutely agree with what you mentioned. I do have a question on the BinTray API for that. If I read the documentation of the API correctly, I'd need to call the endpoint for `GetVersion`. That would be equivalent to calling

```go
versionPath, err := versions.CreatePath("jfrog/jfrog-cli-go/<package>/_latest")
btManager.ShowVersion(versionPath)
```

According to the API docs, it does require an authenticated user and I'm not sure which user that should be as not everyone will have access to that repository. Thoughts are welcome!

eyalbe4: The existing *jfrog-client-go* Bintray API may require authentication, but the Bintray's *get version* API does not.
You can see ti by running
```
curl -XGET https://api.bintray.com/packages/jfrog/jfrog-cli-go/jfrog-cli-linux-amd64/versions/_latest
```
which returns
```
{"name":"1.23.1","desc":null,"package":"jfrog-cli-linux-amd64","repo":"jfrog-cli-go","owner":"jfrog","labels":[],"published":true,"attribute_names":[],"created":"2018-12-26T04:39:03.073Z","updated":"2018-12-26T04:45:17.197Z","released":"","ordinal":51.0,"rating_count":0}
```
So we may need to improve the http client infrastructure to allow non authenticated requests to Bintray.
Let me know if you need any help with this 😄

PR #311, opened 2019-01-13
Create HttpClient using client-builder

barbelity:


eyalbe4:
Awesome @barbelity!  Like!!

PR #316, opened 2019-01-23
Use RetryExecutor in Artifactory and Bintray tests

barbelity:


jfrog-cli/jfrog/artifactory_test.go
```diff
@@ -2191,19 +2192,24 @@ func TestArtifactoryDownloadByArchiveEntriesCli(t *testing.T) {
...
 	artifactoryCli.Exec("upload", "--spec="+uploadSpecFile)
 
-	// Create descriptor for running with retries
-	descriptor := tests.RunWithRetries{
-		Test:            t,
-		FuncToRun:       validateDownloadByArchiveEntries,
-		Expected:        tests.GetBuildArchiveEntriesDownloadCli(),
-		Args:            []string{"dl", tests.Repo1, "out/", "--archive-entries=(*)c1.in", "--flat=true"},
+	// Create executor for running with retries
+	retryExecutor := retry.RetryExecutor{
```
eyalbe4: The use of the retryExecutor in the 3 functions which tests the archives functionality can be moved into a function. The only difference between the 3 retryExecutor definitions seems to be the CLI command arguments (which can be sent to the new function as an argument).

eyalbe4:
👍 @barbelity - See my inline suggestion.

jfrog-cli/jfrog/artifactory_test.go
barbelity: Moved creation of retryExecutor to a function.

PR #330, opened 2019-02-11
Fix collect issues from LATEST version

barbelity:


RobiNino:
Looks good @barbelity

PR #337, opened 2019-02-21
https_proxy fix for download behind a web proxy on CI machines

broekema41:
re-opened see review comments in https://github.com/jfrog/jfrog-cli-go/pull/335

Dear jfrog team.

This fix will allow CI machines that only have access to the internet over a enterprise proxy to use the npm package.

If you have questions please do not hesitate to ask. I will post a temporary npm package on jfrog-cli-go-proxy.

kind regards,
Roland

### DIFF
--- a/npm/init.js
+++ b/npm/init.js
@@ -19,11 +19,13 @@ function validateNpmVersion() {
 
 function download(url) {
     // We detect proxy by looking at the environment variable
+    var agent = https;
     if(process.env.https_proxy && process.env.https_proxy.length > 0) {
+        agent = http;
         url = process.env.https_proxy + starturl.replace('https://', '/https/');
     }
 
-    https.get(url, function(res) {
+    agent.get(url, function(res) {
         if(res.statusCode == 301 || res.statusCode == 302) {
             download(res.headers.location);
         } else if (res.statusCode == 200) {
--- a/npm/init.js
+++ b/npm/init.js
@@ -1,6 +1,7 @@
 validateNpmVersion();
 
-var https = require('follow-redirects').https;
+var https = require('https');
+var http = require('http');
 var fs = require('fs');
 var packageJson = require('./package.json');
 var fileName = getFileName();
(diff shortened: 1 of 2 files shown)


broekema41:
Dear [eyalbe4](/eyalbe4 ) I would love to have this one merged and off my todo list. is there anything i can do?

npm/init.js
```diff
@@ -16,15 +17,30 @@ function validateNpmVersion() {
     }
 }
 
+function download(url) {
+    // We detect proxy by looking at the environment variable
+    var agent = https;
+    if(process.env.https_proxy && process.env.https_proxy.length > 0) {
+        agent = http;
+        url = process.env.https_proxy + starturl.replace('https://', '/https/');
```
eyalbe4: Can you describe this line?
What is the reason for appending the Bintrat URL to the proxy URL?
Shouldn't the Bintray redirection be configured in the proxy server?

npm/init.js
```diff
@@ -16,15 +17,30 @@ function validateNpmVersion() {
     }
 }
 
+function download(url) {
+    // We detect proxy by looking at the environment variable
+    var agent = https;
+    if(process.env.https_proxy && process.env.https_proxy.length > 0) {
```
eyalbe4: Can you please replace
```if(process.env.http...``` with ```if (process.env.http...```
(notice the missing white-space) to remain compatible with the code styling in this file?

File: npm/init.js [hunk 3]
```diff
@@ -16,15 +17,30 @@ function validateNpmVersion() {
...
+        agent = http;
+        url = process.env.https_proxy + starturl.replace('https://', '/https/');
+    }
+
+    agent.get(url, function(res) {
+        if(res.statusCode == 301 || res.statusCode == 302) {
```
eyalbe4: Can you please replace
```if(res.statusCode == 301...``` with ```if (res.statusCode == 301...```
(notice the missing white-space) to remain compatible with the code styling in this file?

File: npm/init.js [hunk 4]
```diff
@@ -16,15 +17,30 @@ function validateNpmVersion() {
...
+            download(res.headers.location);
+        } else if (res.statusCode == 200) {
+            writeToFile(res)
+        } else {
+            console.log('Unexpected status code during JFrog CLI download')
```
eyalbe4: I suggest we add the received status code to the error message.

npm/init.js
broekema41: This way of building the url with https_proxy environment variable is a convention that is used by many applications like curl, wget, jenkins and many more. I decided will make it a bit more fool proof so that it will work in more proxy scenario's.

npm/init.js
broekema41: Done

npm/init.js
broekema41: done

npm/init.js
broekema41: done

npm/init.js
broekema41: I will have a good look at this to see if can think of a better solution.

broekema41: OK. The code for downloading over https proxy is now using a connect tunnel just like curl would do when using --proxy parameter. This way it is no longer needed to manipulate the url and is more according to the standards. 

Additional Info: https://github.com/nodejs/node-v0.x-archive/issues/2474

Hoping you will approve. kind regards, Roland

kevinriemens:
Yes, +1, please accept this change!

PR #351, opened 2019-03-25
Fix bintray access-keys --api-only argument

phyber:
The `jfrog bt access-keys --api-only` argument was not being set correctly and was instead trying to get its state from a non-existant `recursive` flag. This means that any keys created via `jfrog bt access-keys` also allowed access via non-API means.

This PR fixes the issue so that the `--api-only` flag works correctly and the default setting should once again be `--api-only=true` as the `--help` document suggests.

We also add some `vim` files to the `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 .idea
 *.iml
 testsdata/go/project1/go.sum
+
+# vim
+*~
+*.swp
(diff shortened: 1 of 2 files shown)


phyber:
Gentle poke for this CLI argument fix.

eyalbe4:
Thank you for this fix @phyber!
Can you please create the PR agains the *dev* branch as the *Pull Requests Guidelines* section in the README suggest?

phyber:
Hi, sorry about that. I missed that instruction when reading the guideline. The branch has now been rebased against `dev` and the PR is now targetting the `dev` branch.

eyalbe4:
Thanks @phyber!

PR #359, opened 2019-04-07
Support proxy for Gradle deployment

AlexeiVainshtein:


jfrog-cli/artifactory/utils/buildinfoproperties.go
```diff
@@ -193,6 +207,25 @@ func CreateBuildInfoPropertiesFile(buildName, buildNumber string, config *viper.
 	return propertiesFile.Name(), nil
 }
 
+// If HTTP_PROXY env variable presented, adds to the config to deploy Gradle/Maven artifacts via proxy
+func setProxyIfDefined(config *viper.Viper) error {
+	// Add HTTP_PROXY if exists
+	proxy := os.Getenv(HttpProxy)
+	if proxy != "" {
+		url, err := url.Parse(proxy)
+		if err != nil {
```
eyalbe4: CheckError missing.

jfrog-cli/artifactory/utils/buildinfoproperties.go
```diff
@@ -193,6 +207,25 @@ func CreateBuildInfoPropertiesFile(buildName, buildNumber string, config *viper.
...
+			return err
+		}
+		host, port, err := net.SplitHostPort(url.Host)
```
eyalbe4: CheckError missing.

jfrog-cli/artifactory/utils/buildinfoproperties.go
```diff
@@ -193,6 +207,25 @@ func CreateBuildInfoPropertiesFile(buildName, buildNumber string, config *viper.
 	return propertiesFile.Name(), nil
 }
 
+// If HTTP_PROXY env variable presented, adds to the config to deploy Gradle/Maven artifacts via proxy
```
eyalbe4: I suggest rephrasing the comment to:
// If the HTTP_PROXY environment variable is set, add to the config proxy details.

jfrog-cli/artifactory/utils/buildinfoproperties.go
```diff
@@ -193,6 +207,25 @@ func CreateBuildInfoPropertiesFile(buildName, buildNumber string, config *viper.
...
+func setProxyIfDefined(config *viper.Viper) error {
```
eyalbe4: In addition to the HTTP_PROXY env var, we should probably also support the *HTTPS_PROXY* env var.
See the *TestArtifactoryProxy* as a reference.

jfrog-cli/artifactory/utils/buildinfoproperties_test.go
```diff
@@ -37,19 +57,30 @@ func testCreateDefaultPropertiesFile(buildType BuildType, t *testing.T) {
 		}
 	}
 
+	for key, value := range propertiesFileConfig {
```
eyalbe4: Question - why do we need both the * propertiesFileConfig* and *expectedConfig* maps? Aren't they always identical?

jfrog-cli/artifactory/utils/buildinfoproperties_test.go
```diff
@@ -28,6 +38,16 @@ func testCreateDefaultPropertiesFile(buildType BuildType, t *testing.T) {
 		t.Error(err)
 	}
 
+	var yamlConfig = map[string]string{
```
eyalbe4: Is the *yamlConfig* map necessary?

jfrog-cli/artifactory/utils/buildinfoproperties_test.go
```diff
@@ -134,3 +166,32 @@ func compareViperConfigs(t *testing.T, actual, expected *viper.Viper, buildType
 		}
 	}
 }
+
+func TestSetProxyIfNeeded(t *testing.T) {
```
eyalbe4: Question: Isn't this test sufficient for testing the proxy functionality? If it is, can we remove the new additions to the other tests in this file?

jfrog-cli/artifactory/utils/buildinfoproperties.go
AlexeiVainshtein: We don't support https in the build info. We only support http. I've checked and the proxy is usually running on http only without https.

jfrog-cli/artifactory/utils/buildinfoproperties_test.go
AlexeiVainshtein: I think that this only test the setProxyIfDefined function. The other tests testing the entire viper that the information is written correctly in the form of artifactory.proxy.host and artifactory.proxy.port. Those are the values that the build info needs so the other tests purpose is to check that those are the correct keys.

jfrog-cli/artifactory/utils/buildinfoproperties_test.go
AlexeiVainshtein: I removed it

jfrog-cli/artifactory/utils/buildinfoproperties_test.go
AlexeiVainshtein: Not exactly. The expectedConfig is a Viper not a map. After creating the viper and populating the viper information we adding the propertiesFileConfig key and value to the viper. We can change this to:

	expectedConfig.Set("artifactory.proxy.host", host)
	expectedConfig.Set("artifactory.proxy.port", port)

jfrog-cli/artifactory/utils/buildinfoproperties.go
AlexeiVainshtein: Done

jfrog-cli/artifactory/utils/buildinfoproperties.go
AlexeiVainshtein: Done

jfrog-cli/artifactory/utils/buildinfoproperties.go
AlexeiVainshtein: Done

jfrog-cli/artifactory/utils/buildinfoproperties_test.go
AlexeiVainshtein: I've changed it per our conversation.

PR #360, opened 2019-04-08
If url without trailing slash then rt config command with access token fails

AlexeiVainshtein:


jfrog-cli/artifactory/commands/config.go
```diff
@@ -136,6 +135,7 @@ func getConfigurationFromUser(details, defaultDetails *config.ArtifactoryDetails
 	if fileutils.IsSshUrl(details.Url) {
 		return getSshKeyPath(details)
 	}
+	details.Url = clientutils.AddTrailingSlashIfNeeded(details.Url)
```
eyalbe4: I would suggest moving the addition of the trailing slash to the *prepareConfigurationData* function.
Here's the reason why:
With this change, the addition of the slash is done in the *getConfigurationFromUser* function. This function however gets the *details* by value and not by reference. This means that the change in the URL does not affect th4e details struct outside of the method.
Although this change probably fixes the bug, if in the future the *details* struct will be used again outside of the function, the URL will be without the trailing slash.

AlexeiVainshtein: We can't. In the getConfigurationFromUser --> line 142 --> readAccessTokenFromConsole --> we perform ping to Artifactory. If the user adds a URL from the console in line 132 and the URL without the trailing slash, then the ping fails. Also, we do pass the pointer *config.ArtifactoryDetails so the url that being changed here has been also changed outside the scope of the function as well (I've tested it) so the URL has a trailing slash afterwards.

PR #361, opened 2019-04-09
Introduce generic progress bar

RobiNino:
Introduce a new progress bar to upload/download to/from artifactory


utils/log/progress.go
```diff
@@ -0,0 +1,292 @@
...
+	return newProgressBar, logFile, nil
+}
+
+// Init progress bar if all required conditions are met:
+// CI == false (or unset), Stderr is a terminal, and terminal width is large enough
+func shouldInitProgressBar() (bool, error) {
+	isCI, err := utils.GetBoolEnvValue("CI", false)
+	if err != nil {
+		return false, err
+	}
```
eyalbe4: In case isCI is false, we can probably return here and save executing the upcoming code.

utils/log/progress.go
```diff
@@ -0,0 +1,292 @@
...
+	newProgressBar.printLogFilePathAsBar(logFile.Name())
+	newProgressBar.newHeadlineBar(" Working... ")
+
+	return newProgressBar, logFile, nil
+}
+
+// Init progress bar if all required conditions are met:
+// CI == false (or unset), Stderr is a terminal, and terminal width is large enough
+func shouldInitProgressBar() (bool, error) {
+	isCI, err := utils.GetBoolEnvValue("CI", false)
```
eyalbe4: It is better to use **ci** instead of **isCI** according to the common code conversions.

utils/log/progress.go
```diff
@@ -0,0 +1,292 @@
...
+	// Bar replacement is a spinner and thus does not use a channel for incrementing
+	unit := &progressBarUnit{bar: newBar, incrChannel: nil, replaceBar: p.bars[replaceBarId-1].bar}
+	// Add bar to bars array
+	p.bars = append(p.bars, unit)
+	barId = len(p.bars)
+	p.barsRWMutex.Unlock()
+	return barId
+}
+
+func buildProgressDescription(prefix, path string, extraCharsLen int) string {
```
eyalbe4: I suggest adding a unit test for this function.

utils/log/log.go
```diff
@@ -0,0 +1,57 @@
...
+	file, err := os.OpenFile(fileName, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
+	if err != nil {
+		return nil, errorutils.CheckError(err)
+	}
+
+	return file, nil
+}
+
+// Closes the log file and resets to the default logger
+func CloseLogFile(logFile *os.File) {
```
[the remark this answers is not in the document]
RobiNino: I think it is good practice to close the file manually and report errors that might pop only when closing the file.
Since we are closing the log file, setting the logger to default is necessary if we want to log the error if one does occur.

PR #374, opened 2019-05-05
 Support access-token in Artifactory commands

barbelity:


artifactory/commands/curl/curl.go
```diff
@@ -48,15 +48,29 @@ func Execute(args []string) error {
...
+	cmdWithoutCreds := strings.Join(command.Arguments, " ")
+	// Add credentials to curl command.
+	credentialsMessage, err := command.addCommandCredentials(artDetails)
 
 	// Run curl.
+	log.Debug(fmt.Sprintf("Executing curl command: '%s %s'", cmdWithoutCreds, credentialsMessage))
 	return gofrogcmd.RunCmd(command)
 }
 
+func (curlCmd *CurlCommand) addCommandCredentials(artDetails *config.ArtifactoryDetails) (string, error) {
```
eyalbe4: This function seems to never return an error, so we may need to remove the error as a returned value.

PR #380, opened 2019-05-12
Added --skip-login flag to docker-pull and docker-push.

eyalbe4:


djgilcrease:
This looks like a better option then implicitly skipping as my PR does https://github.com/jfrog/jfrog-cli-go/pull/353 feel free to close the other PR

artifactory/cli.go
```diff
@@ -680,6 +680,13 @@ func getBuildToolFlags() []cli.Flag {
 	}
 }
 
+func getSkipLoginFlag() cli.Flag {
```
AlexeiVainshtein: Both docker push and docker pull are using the same flags such as:
getBuildToolFlags, getServerFlags and getSkipLoginFlag. The only difference is in the docker push command that has also the getThreadsFlag. Perhaps we should create a getDockerFlags that will include the getBuildToolFlags, getServerFlags and getSkipLoginFlag and for docker push also add the threads flags:

```
func getDockerFlags() []cli.Flag {
	var flags []cli.Flag
	flags = append(flags, getBuildToolFlags()...)
	flags = append(flags, getServerFlags()...)
	flags = append(flags, getSkipLoginFlag())
	return flags
}

func getDockerPushFlags() []cli.Flag {
	var flags []cli.Flag
	flags = append(flags, getDockerFlags()...)
	flags = append(flags, getThreadsFlag())
	return flags
}
```

artifactory/cli.go
```diff
@@ -1324,10 +1333,12 @@ func dockerPushCmd(c *cli.Context) {
 	artDetails := createArtifactoryDetailsByFlags(c, true)
 	imageTag := c.Args().Get(0)
 	targetRepo := c.Args().Get(1)
+		skipLogin := c.Bool("skip-login")
+
```
AlexeiVainshtein: Redundant white spaces

artifactory/commands/docker/pull.go
```diff
@@ -24,10 +24,12 @@ func (dpc *DockerPullCommand) Run() error {
 	if errorutils.CheckError(err) != nil {
 		return err
 	}
-	loginConfig := &docker.DockerLoginConfig{ArtifactoryDetails: rtDetails}
-	err = docker.DockerLogin(imageTag, loginConfig)
-	if err != nil {
-		return err
+	if !dpc.skipLogin {
+		loginConfig := &docker.DockerLoginConfig{ArtifactoryDetails: rtDetails}
+		err = docker.DockerLogin(imageTag, loginConfig)
```
AlexeiVainshtein: I think we can remove the imageTag variable and just use the dpc.ImageTag() here. So the `	imageTag := dpc.ImageTag()` will be removed and instead as in the push command all will be done on the dpc.ImageTag()

artifactory/commands/docker/push.go
```diff
@@ -33,18 +33,20 @@ func (dpc *DockerPushCommand) Run() error {
...
 		return err
 	}
-	loginConfig := &docker.DockerLoginConfig{ArtifactoryDetails: rtDetails}
-	imageTag := dpc.ImageTag()
-	err = docker.DockerLogin(imageTag, loginConfig)
-	if err != nil {
-		return err
+	if !dpc.skipLogin {
+		loginConfig := &docker.DockerLoginConfig{ArtifactoryDetails: rtDetails}
+		imageTag := dpc.ImageTag()
```
AlexeiVainshtein: I think this can be removed and just to use the dpc.ImageTag()

eyalbe4:
Implemented all comments in https://github.com/jfrog/jfrog-cli-go/pull/388
I'm therefore closing this one.

fdolsky321:
Guys, is there a way to push docker image to artifactory, when the config.json file is restricted to write (by admins). I was trying to use --user and --api-key with --skip-login, but then I was not authorized at all and in artifactory log, I saw HTTP request from anonymous rather then the user which was used in --user. Do you have any idea, is this possible? @eyalbe4 @AlexeiVainshtein 
Thanks a lot.

PR #383, opened 2019-05-19
Support Unix Syntax on Windows

RobiNino:


artifactory_test.go
```diff
@@ -2921,6 +2923,34 @@ func TestArtifactoryBuildDiscard(t *testing.T) {
 	cleanArtifactoryTest()
 }
 
+func TestArtifactoryWinBackwardsCompatibility(t *testing.T) {
```
eyalbe4: I suggest adding a description comment to this function.

bintray_test.go
```diff
@@ -329,6 +328,38 @@ func TestBintrayVersionDownloads(t *testing.T) {
 	cleanBintrayTest()
 }
 
+func TestBintrayUploadWindowsCompatibility(t *testing.T) {
```
eyalbe4: I suggest adding a description comment to this function.

PR #386, opened 2019-05-26
Support Go native and go-config commands

AlexeiVainshtein:


artifactory/commands/golang/gorecursivepublish.go
```diff
@@ -93,11 +101,16 @@ func (grp *GoRecursivePublishCommand) Run() error {
 			return err
 		}
 	}
-	err = gocmd.RecursivePublish(grp.TargetRepo(), goModEditMessage, serviceManager)
+
+	goInfo := &params.ResolverDeployer{}
+	deployerResolver := &params.Params{}
```
eyalbe4: deployerResolver -- > resolverDeployer

artifactory/utils/argsutils.go
```diff
@@ -0,0 +1,91 @@
...
+	argsCopy := *args
+	// Remove flag from Command if required.
+	if flagIndex != -1 {
+		argsCopy = append(argsCopy[:flagIndex], argsCopy[flagValueIndex+1:]...)
+	}
+	*args = argsCopy
+}
+
+// Find value of required CLI flag in Command.
+// If flag does not exist, returned indexes are -1 and error is nil.
```
eyalbe4: ...returned indexes are -1  and error is nil. --> ...the returned index is -1 and nil is returned as the error.

artifactory/cli.go
```diff
@@ -484,6 +498,15 @@ func GetCommands() []cli.Command {
 	}
 }
 
+func getGlobalConfigFlag() []cli.Flag {
+	return []cli.Flag{
+		cli.BoolFlag{
+			Name:  "global",
+			Usage: "[Default: false] Saves the config file to JFROG_CLI_HOME_DIR.` `",
```
eyalbe4: Saves the config file to JFROG_CLI_HOME_DIR. --> Set to true, if you'd like to configuration to be global (for all projects). Specific projects can override the global configuration.

artifactory/cli.go
```diff
@@ -1235,6 +1263,19 @@ func validateServerId(serverId string) {
 	}
 }
 
+func validateGoNativeCommand(args []string) string {
```
eyalbe4: I suggest adding a description comment for this function.
I think that it'll be better if this function actually returns an error, with the invalid flag it finds. It may make the function (and also its test) easier to understand.

buildtools_test.go
```diff
@@ -308,6 +311,64 @@ func TestGoBuildInfo(t *testing.T) {
...
+	// 2. Publish build-info.
+	// 3. Validate the total count of dependencies added to the build-info.
+	buildNumber := "1"
+	// Preparing config file.
+	configFileDir := filepath.Join(project1Path, ".jfrog", "projects")
+	configFileDir, err = tests.ReplaceTemplateVariables(filepath.Join(configFileDir, "go.yaml"), configFileDir)
+	if err != nil {
+		t.Error(err)
+	}
+	artifactoryGoCli := tests.NewJfrogCli(execMain, "jfrog rt", "")
```
eyalbe4: What is the reason for creating a new JFrog CLI here, instead of using the existing one?

eyalbe4:
Awesome work @AlexeiVainshtein!
See my inline comments.

artifactory/commands/golang/goconfig.go
```diff
@@ -0,0 +1,92 @@
...
+		err = configResult.Deployer.Server.Set(vConfig)
+		if err != nil {
+			return err
+		}
+		availableRepos, err := prompt.GetRepositories(vConfig, utils.LOCAL, utils.VIRTUAL)
+		if err != nil {
+			// If there are no available repos pass empty array.
+			availableRepos = []string{}
+		}
+		configResult.Deployer.Repo, err = prompt.ReadRepo("Set repository for dependency deployment (press Tab for options): ", availableRepos)
```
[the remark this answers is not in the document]
AlexeiVainshtein: Done

artifactory/commands/golang/gorecursivepublish.go
AlexeiVainshtein: Done

artifactory/cli.go
AlexeiVainshtein: Done. Should it be:  Set to true, if you'd like **the** configuration to be global (for all projects). Specific projects can override the global configuration. ?

buildtools_test.go
AlexeiVainshtein: Since the existing one always send the flags of --url, --user, --password. Those flags can't be used with rt go and config file.

artifactory/utils/argsutils.go
AlexeiVainshtein: Done

artifactory/cli.go

PR #387, opened 2019-05-27
Support for publishing info file for Go

AlexeiVainshtein:


artifactory/utils/golang/project/project.go
```diff
@@ -102,10 +118,57 @@ func (project *goProject) PublishPackage(targetRepo, buildName, buildNumber stri
 	if err != nil {
 		return err
 	}
+	// Create the info file if Artifactory version is 6.10.0 and above.
+	version := version.NewVersion(_go.ArtifactoryMinSupportedVersionForInfoFile)
+	artifactoryVersion, err := servicesManager.GetConfig().GetArtDetails().GetVersion()
+	if err != nil {
+		return err
+	}
+	if version.AtLeast(artifactoryVersion) {
```
eyalbe4: I think that the above if statement means - 
I the min-supported-version is at least the Artifactory version - it seems like it should be the other way around.
It is worth also checking the other info file version comparisons.

artifactory/utils/golang/project/project.go
```diff
@@ -102,10 +118,57 @@ func (project *goProject) PublishPackage(targetRepo, buildName, buildNumber stri
...
+		}
+		params.InfoPath = pathToInfo
+	}
 
 	return servicesManager.PublishGoProject(params)
 }
 
+// Creates the info file.
+// Returns the path to that file.
+func (project *goProject) createInfoFile() (string, error) {
```
eyalbe4: This function has a few CheckError calls missing.

artifactory/utils/golang/project/project.go
```diff
@@ -226,6 +289,19 @@ func (project *goProject) archiveProject(version, tempDir string) (string, error
...
 
+// Add the info file also as an artifact to be part of the build info.
+func (project *goProject) addInfoFileToBuildInfo(infoFilePath string) error {
+	fileDetails, err := fileutils.GetFileDetails(infoFilePath)
+	if err != nil {
+		return err
+	}
+
+	artifact := buildinfo.Artifact{Name: project.version + ".info"}
+	artifact.Checksum = &buildinfo.Checksum{Sha1: fileDetails.Checksum.Sha1, Md5: fileDetails.Checksum.Md5}
```
eyalbe4: Question - Haven't we already calculated checksum for the info file when we published it to Artifactory?
If so, we should reuse the calculated checksum and not calculate it again.

artifactory/utils/golang/project/project.go
AlexeiVainshtein: Per our discussion, I've changed the api.

artifactory/utils/golang/project/project.go
AlexeiVainshtein: Done

artifactory/utils/golang/project/project.go
AlexeiVainshtein: No. The publish to Artifactory is using api/go and therefore we don't calculate the checksum. The checksum is being calculated once at this stage. We can't use checksum deploy since the url is build with api/go

PR #390, opened 2019-05-29
Use global configuration if local are missing for go

AlexeiVainshtein:


artifactory/cli.go
```diff
@@ -1486,10 +1485,16 @@ func shouldSkipGoFlagParsing() bool {
 	}
 
 	if !exists {
-		return false
+		// If missing in the root project, check in the home dir
+		jfrogRootDir, err = config.GetJfrogHomeDir()
```
eyalbe4: I think that the name of the * jfrogRootDir* variable should be changed, because it is actually not the root dir.
What do you think?
How about *ProjectConfPath*?

artifactory/cli.go
```diff
@@ -1486,10 +1485,16 @@ func shouldSkipGoFlagParsing() bool {
...
+		if err != nil {
+			cliutils.ExitOnErr(err)
+		}
+	} else {
+		jfrogRootDir = filepath.Join(jfrogRootDir, ".jfrog")
 	}
 
-	configFilePath := filepath.Join(jfrogRootDir, ".jfrog", "projects", "go.yaml")
+	configFilePath := filepath.Join(jfrogRootDir, "projects", "go.yaml")
```
eyalbe4: It looks like there a few places in the code which construct the **<project dir>/project/go.yaml* similarly. Maybe we should consolidate all of them into one function.
In addition, notice that the logic of finding the go.yaml file (searching for it upstream and then in the home dir) is duplicated in the *goCmd* and *shouldSkipGoFlagParsing* functions.

AlexeiVainshtein: Done I've created one function that both goCmd and shouldSkipGoFlagParsing are using

artifactory/cli.go
AlexeiVainshtein: This one I think should be jfrogHomeDir since it represents the home dir. The jfrogRootDir above this I renamed to projectDir. Let me know what you think.

PR #394, opened 2019-06-06
Add support for npm ci command

AlexeiVainshtein:


artifactory/cli.go
```diff
@@ -777,6 +790,10 @@ func getNpmCommonFlags() []cli.Flag {
 }
 
 func getNpmInstallFlags() []cli.Flag {
```
eyalbe4: I think that the *getNpmInstallFlags* and *getNpmCiFlags* functions can be removed.
I think that both the *npm install* and *npm ci* commands can use the *getNpmFlags* function.

PR #399, opened 2019-06-16
Fix set properties deleting properties instead of set

AlexeiVainshtein:


artifactory_test.go
```diff
@@ -874,15 +880,18 @@ func TestArtifactorySetPropertiesExcludeByCli(t *testing.T) {
...
+		t.Error("No artifacts were found.")
+	}
 
 	for _, item := range resultItems {
 		properties := item.Properties
 		for _, prop := range properties {
-			if prop.Key != "" {
-				t.Error("Wrong properties")
+			if prop.Key == "prop" {
+				t.Error("Property prop were not deleted from artifact", item.Name)
```
eyalbe4: Property prop were not deleted from artifact --> Property 'prop' was not deleted from artifact

AlexeiVainshtein: Done

PR #402, opened 2019-06-20
Fix spelling mistake in npm/install.go log

Lykathia:
Noticed while setting up / debugging some Jenkins automation. Must have been overlooked in PR review previously.
- - -
_(OT: from debugging, the jfrog CLI makes too many individual calls when dealing w/ NPM artifacts, resulting in my jenkins server getting banned off the network w/ default IT firewalls ... seems like a really silly overhead to have ... this seems to be by design, so I'm not sure if it's worth opening it's own bug or not)_

### DIFF
--- a/artifactory/commands/npm/install.go
+++ b/artifactory/commands/npm/install.go
@@ -291,7 +291,7 @@ func (nca *NpmCommandArgs) setDependenciesList() (err error) {
 }
 
 func (nca *NpmCommandArgs) collectDependenciesChecksums() error {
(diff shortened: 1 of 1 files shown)


Lykathia:
Note there is another mistake upstream [here](https://github.com/jfrog/gofrog/blob/master/parallel/runner.go#L70), which makes reading the code a bit tricky.

eyalbe4:
Thank you @Lykathia!

PR #409, opened 2019-06-27
 Support bash and zsh auto-completion 

yahavi:


artifactory/commands/config.go
```diff
@@ -415,3 +415,15 @@ type ConfigCommandConfiguration struct {
 	Interactive bool
 	EncPassword bool
 }
+
+func GetAllServerIds() []string {
```
eyalbe4: I suggest moving this function to **artifactory/utils/config/config.go** and renaming it to *GetAllArtifactoryServerIds*.

PR #411, opened 2019-07-01
Change the API npm uses for resolving Artifactory version

barbelity:


artifactory/commands/npm/install.go
```diff
@@ -610,13 +610,13 @@ func getDetailsUsingBasicAuth(artDetails auth.ArtifactoryDetails) (npmAuth strin
 		return "", "", errorutils.CheckError(errors.New("Artifactory response: " + resp.Status + "\n" + cliutils.IndentJson(body)))
 	}
 
-	// Extract Artifactory version from response header.
-	serverValues := strings.Split(resp.Header.Get("Server"), "/")
-	if len(serverValues) != 2 {
-		errorutils.CheckError(errors.New("Cannot parse Artifactory version from the server header."))
+	// Get Artifactory version.
```
eyalbe4: Now that the we use a separate API for getting the Artifactory version, it is probably better to invoke it first, validate the version and only if the validation passes, get the npmAuth from Artifactory.
This is relevant to two functions - *getDetailsUsingBasicAuth* and *getDetailsUsingAccessToken*.

barbelity: Added an Artifactory version check before fetching the npm token.

PR #412, opened 2019-07-02
Add count flag for search command

AlexeiVainshtein:


artifactory/cli.go
```diff
@@ -952,6 +952,10 @@ func getSearchFlags() []cli.Flag {
 			Name:  "build",
 			Usage: "[Optional] If specified, only artifacts of the specified build are matched. The property format is build-name/build-number. If you do not specify the build number, the artifacts are filtered by the latest build number.` `",
 		},
+		cli.BoolFlag{
+			Name:  "count",
+			Usage: "[Optional] Set to true if you do not wish to print the entire JSon result but only the total of found artifacts.` `",
```
eyalbe4: I suggest changing the description to:
*Set to true to display only the total of files or folders found.*

artifactory/cli.go
```diff
@@ -1771,7 +1775,9 @@ func searchCmd(c *cli.Context) {
 	cliutils.ExitOnErr(err)
 	result, err := json.Marshal(searchCmd.SearchResult())
 	cliutils.FailNoOp(err, len(searchCmd.SearchResult()), 0, isFailNoOp(c))
-	log.Output(string(clientutils.IndentJson(result)))
+	if !c.Bool("count") {
```
eyalbe4: Let's changes this, so that the files / folders count is displayed.

artifactory/cli.go

artifactory/cli.go

PR #413, opened 2019-07-03
Support include-dirs flag on search

yahavi:


artifactory/commands/generic/search.go
```diff
@@ -11,12 +11,14 @@ import (
```
eyalbe4: To remain consistent with the upload and download commands, I think that in the includeDirs property should also be added to a File Spec.

artifactory/cli.go
```diff
@@ -952,6 +952,10 @@ func getSearchFlags() []cli.Flag {
 			Name:  "build",
 			Usage: "[Optional] If specified, only artifacts of the specified build are matched. The property format is build-name/build-number. If you do not specify the build number, the artifacts are filtered by the latest build number.` `",
 		},
+		cli.BoolFlag{
```
eyalbe4: Since this option is shared now between two commands, I suggest using a new *getIncludeDirsFlag* function.

PR #415, opened 2019-07-03
Adding Diagnosis command to cli

dashbhanu:
A new features in CLI to get a diagnosis of Artifactory after each upgrade. This command should run before and after upgrade, so that it will print the performance and stability of artifactory instance. The default command is like
jfrog  rt diagnosis

### DIFF
--- a/artifactory/cli.go
+++ b/artifactory/cli.go
@@ -51,6 +51,7 @@ import (
 	nugetdocs "github.com/jfrog/jfrog-cli-go/docs/artifactory/nuget"
 	nugettree "github.com/jfrog/jfrog-cli-go/docs/artifactory/nugetdepstree"
 	"github.com/jfrog/jfrog-cli-go/docs/artifactory/ping"
+	"github.com/jfrog/jfrog-cli-go/docs/artifactory/diagnosis"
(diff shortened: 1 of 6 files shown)


eyalbe4:
@dashbhanu,
First, thanks for working on this!

I have to say that I'm having issues with the basic design of this functionality. There's a lot I think we need to modify, so I'm starting from some of the most significant points:

Reading plain text (not formatted) response and then parsing it manually, is something we really want to avoid.
It is okay receive a JSON responses and then reading it into a struct, but parsing strings is not the way to go from our perspective. In addition, the content is currently saved to files as a stream of bytes, which make it unreadable.

I don't think that the code should be included as part of the jfrog-client-go library, because I find it hard it believe that it is useful for other applications - at least they way the API is built now.

There's a lot more to discuss, in terms of where the info files should be saved to, how the command should receive them, code styling and more, but I suggest we start with the above first.

Thanks

dashbhanu:
Thank you @eyalbe4  for your time, I will take some more time and go throw your review comments and then again resubmit PR.

Thanks

eyalbe4:
@dashbhanu,
Would you like to continue working on this?

eyalbe4:
Closing this one for now.

PR #422, opened 2019-07-28
Fix dependency-id for Nuget projects with project.assets.json file

barbelity:


artifactory/utils/nuget/dependencies/assetsjson.go
```diff
@@ -142,6 +142,16 @@ func (assets *assets) isPackagePartOfTargetDependencies(nugetPackageName string)
 	return false
 }
 
+// Dependencies-id in assets is built in form of: <package-name>/<version>.
+// The Build-info format of dependency id is: <package-name>:<version>.
+func getDependencyIdForBuildInfo(dependencyAssetId string) string {
+	indexOfSlash := strings.Index(dependencyAssetId, "/")
```
yahavi: Why not using [replace](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#replace(char,%20char))/[replaceFirst](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#replaceFirst(java.lang.String,%20java.lang.String))?

PR #427, opened 2019-08-02
ARM CLI Support

btashton:
Add ARM support to the cli build jobs as well as getCli script.

Linux packages have been uploaded and tested on 32bit and 64bit arm here: https://bintray.com/beta/#/btashton/jfrog-cli-go?tab=packages

Does not yet add building ARM docker images.


eyalbe4:
Awesome @btashton! Thanks!

PR #434, opened 2019-08-14
Removed an invalid regexp escape "\a"

steveonjava:
The bell character (\a) causes this test to fail when run against Windows artifactory servers.


steveonjava:
The bell character (\a) causes this test to fail when run against Windows artifactory servers.

RobiNino:
Thanks @steveonjava !

PR #443, opened 2019-08-25
Artifactory download - Add flag for sync-deletes

liron-shalom:


utils/tests/utils.go
```diff
@@ -130,6 +141,21 @@ func compare(expected, actual []string) error {
...
 
+// Actual list contains dir and files, so we check prefix (for dirs) or equality (for files)
+func isExclusivelyExistLocally(expected, actual []string) error {
+	for _, v := range actual {
+		for i, r := range expected {
+			if strings.HasPrefix(r, v) || v == r {
+				break
+			}
+			if i == len(actual)-1 {
+				return errors.New("Should not have : " + v)
```
yahavi: Following our discussion -
Add t parameter to isExclusivelyExistLocally and add t.Error. Don't return err.
```suggestion
				t.Error("Should not have : " + v)
```

utils/tests/utils.go
```diff
@@ -82,6 +82,17 @@ func removeDirs(dirs ...string) {
 	}
 }
 
+// After syncDeletes we must make sure that there are no actual files we did not expect.
+func CheckSyncedDirContent(expected, actual []string, t *testing.T) {
+	// IsExistLocally - check if all expected are actually exists
```
eyalbe4: // IsExistLocally - check if all expected are actually exists -->
// Check if all expected actually exist.

utils/tests/utils.go
```diff
@@ -82,6 +82,17 @@ func removeDirs(dirs ...string) {
...
+	IsExistLocally(expected, actual, t)
+	// IsExclusivelyExistLocally - check if we expected all the actual
```
eyalbe4: Not sure how clear is above comment...

utils/tests/utils.go
```diff
@@ -130,6 +141,21 @@ func compare(expected, actual []string) error {
 	return nil
 }
 
+// Actual list contains dir and files, so we check prefix (for dirs) or equality (for files)
```
eyalbe4: Let's try to improve this comment. I'm pretty sure it won't be clear for the random user.

utils/tests/utils.go
```diff
@@ -130,6 +141,21 @@ func compare(expected, actual []string) error {
...
+func isExclusivelyExistLocally(expected, actual []string, t *testing.T) error {
```
eyalbe4: This function returns an error, but the error returned is always nil.

utils/tests/utils.go
```diff
@@ -82,6 +82,17 @@ func removeDirs(dirs ...string) {
 	}
 }
 
+// After syncDeletes we must make sure that there are no actual files we did not expect.
+func CheckSyncedDirContent(expected, actual []string, t *testing.T) {
```
eyalbe4: Looks like this function is used only from *artifactory_test.go*. Any reason why not move it there?
It is very specific for the Artifactory upload and download commands. It is currently inside a generic test util file.

artifactory/commands/generic/download.go
```diff
@@ -104,6 +108,12 @@ func (dc *DownloadCommand) Run() error {
 		dc.result.SetSuccessCount(totalExpected)
 		dc.result.SetFailCount(0)
 		return err
+	} else if dc.SyncDeletesPath() != "" {
```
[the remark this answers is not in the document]
liron-shalom: I wanted to prevent running dry-run+sync-deletes - 
so if dry-run is set, we will skip the sync-deletes logic.

PR #447, opened 2019-09-02
Artifactory tests : Run UploadDebian test with separate repository

liron-shalom:


testsdata/debian_test_repository_config.json
```diff
@@ -0,0 +1,37 @@
+{
+  "key": "${DEBIAN_REPO}",
+  "packageType": "generic",
```
yahavi: Shouldn't be "debian"?

PR #450, opened 2019-09-05
Artifactory, Maven - Add MAVEN_OPTS env var to mvn command

liron-shalom:


artifactory/commands/mvn/mvn.go
```diff
@@ -156,6 +156,11 @@ func (mc *MvnCommand) createMvnRunConfig(dependenciesPath string) (*mvnRunConfig
 		return nil, errorutils.CheckError(err)
 	}
 
+	mavenOpts := os.Getenv("MAVEN_OPTS")
+	if mavenOpts != "" {
+		mavenOpts = strings.ReplaceAll(mavenOpts, "\n\t\r", " ")
```
eyalbe4: I think that the replace is redundant, because it's unlikely that an environment variable includes these characters.

PR #452, opened 2019-09-08
Add import and export server commands

yahavi:


docs/artifactory/config/help.go
```diff
@@ -13,6 +15,12 @@ const Arguments string = `	server ID
 	show
 		Shows the stored configuration.
 		In case this argument is followed by a configured server ID, then only this server's configurations is shown.
+	
+	import
+		Imports an artifactory server configuration from server token.
```
eyalbe4: Imports an artifactory server configuration from server token. -->
Imports an Artifactory server configuration from a server token. A server token is generated by the "jfrog rt c export [Server ID]" command.

docs/artifactory/config/help.go
```diff
@@ -13,6 +15,12 @@ const Arguments string = `	server ID
...
+	
+	export
+		Creates an Artifactory server configuration token.
```
eyalbe4: Creates an Artifactory server configuration token. -->
Creates an Artifactory server configuration token. The generated token can be imported by the "jfrog rt c import [Server token]" command.

PR #453, opened 2019-09-09
Allow buildname, buildnumber, excludeEnv and buildUrl from env 

yahavi:


docs/common/help.go
```diff
@@ -21,6 +21,19 @@ const GlobalEnvVars string = `	JFROG_CLI_LOG_LEVEL
 		[Default: The operating system's temp directory]
 		Defines the temp directory used by JFrog CLI.
 
+	JFROG_CLI_BUILD_NAME
+		Build name to use in build related commands. This environment variable will be used in case of omitted --build-name flag and <build name> argument.
```
eyalbe4: I suggest rephrasing this description to:
Build name to be used by commands which expect a build name, unless sent as a command argument or option.

docs/common/help.go
```diff
@@ -21,6 +21,19 @@ const GlobalEnvVars string = `	JFROG_CLI_LOG_LEVEL
...
+	
+	JFROG_CLI_BUILD_NUMBER
+		Build number to use in build related commands. This environment variable will be used in case of omitted '--build-number' flag and '<build number>' argument.
```
eyalbe4: I suggest rephrasing this description to:
Build number to be used by commands which expect a build number, unless sent as a command argument or option.

docs/common/help.go
```diff
@@ -21,6 +21,19 @@ const GlobalEnvVars string = `	JFROG_CLI_LOG_LEVEL
...
+
+	JFROG_CLI_BUILD_URL
+		Can be used for setting the CI server build URL in the build-info. This environment variable will be used in case of omitted '--build-url' flag.
```
eyalbe4: I suggest rephrasing to:
Sets the CI server build URL in the build-info. The "jfrog rt build-publish" command uses the value of this environment variable, unless the --build-url command option is sent.

docs/common/help.go
```diff
@@ -21,6 +21,19 @@ const GlobalEnvVars string = `	JFROG_CLI_LOG_LEVEL
...
+	
+	JFROG_CLI_ENV_EXCLUDE
+		[Default: *password*;*secret*;*key*;*token*] 
+		List of case insensitive patterns in the form of "value1;value2;...". Environment variables match those patterns will be excluded. This environment variable will be used in case of omitted '--env-exclude' flag.
```
eyalbe4: ...This environment variable will be used in case of omitted '--env-exclude' flag. -->
...This environment variable is used by the "jfrog rt build-publish" command,  in case the --env-exclude command option is not sent.

buildinfo_test.go
```diff
@@ -223,6 +223,8 @@ func TestArtifactoryPublishBuildInfoBuildUrl(t *testing.T) {
 	initArtifactoryTest(t)
```
eyalbe4: This test is pretty good. I suggest we also test that if a build name and build number are set by env vars (and not by command options), the env vars are used.

buildtools_test.go
```diff
@@ -353,6 +353,103 @@ func TestGoBuildInfo(t *testing.T) {
 	cleanGoTest(gopath)
 }
 
+func TestGoBuildnameAndNumberFromEnv(t *testing.T) {
```
eyalbe4: TestGoBuildnameAndNumberFromEnv --> TestGoBuildNameAndNumberFromEnv

Can we reuse the code in the *TestGoBuildInfo* test? I think we should avoid this code duplication.

buildinfo_test.go
yahavi: This is already tested and in 2 levels:
Unit test: TestExtractBuildDetailsFromEnv
Integration test: TestArtifactoryGenericBuildnameAndNumberFromEnv

buildtools_test.go
yahavi: I deleted the test because the native commands flow is already tested in [unit test](https://github.com/jfrog/jfrog-cli/pull/453/files#diff-3f73e7bf71fef561eb5790dc08c1c68bR161).

PR #456, opened 2019-09-18
Add pip tests

barbelity:


utils/tests/utils.go
```diff
@@ -322,12 +326,17 @@ func GetNonVirtualRepositories() map[string]string {
...
+	}
+
 	return nonVirtualRepos
 }
 
 func GetVirtualRepositories() map[string]string {
 	return map[string]string{
-		VirtualRepo: VirtualRepositoryConfig,
+		VirtualRepo:     VirtualRepositoryConfig,
+		PypiVirtualRepo: PypiVirtualRepositoryConfig,
```
yahavi: PypiVirtualRepo should be added only if `*TestPip is true`

PR #461, opened 2019-09-23
Artifactory search : Add size, created, modified to search json result

Postyy:


utils/tests/searchresultnodate.go
```diff
@@ -0,0 +1,110 @@
...
+//Inorder to compare hard coded json's struction result(e.g. GetSearchPropsStep)
+//witch cannot contain future 'Created' & 'Modified' dates
+type SearchResultNoDate struct {
+	Path  string              `json:"path,omitempty"`
+	Type  string              `json:"type,omitempty"`
+	Size  int64               `json:"size,omitempty"`
+	Props map[string][]string `json:"props,omitempty"`
+}
+
+type SearchCommandNoDate struct {
```
eyalbe4: There's probably no reason to duplicate the whole search command. It is enough to add a method on the *SearchResult* struct, which returns a new *SearchResultNoDate* struct.

utils/tests/searchresultnodate.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/jfrog/jfrog-cli-go/artifactory/utils"
+	"github.com/jfrog/jfrog-client-go/artifactory/services"
+	clientutils "github.com/jfrog/jfrog-client-go/artifactory/services/utils"
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+)
+
+//Used in tests TestArtifactorySearchIncludeDir & 'TestArtifactorySearchProps'.
+//Inorder to compare hard coded json's struction result(e.g. GetSearchPropsStep)
+//witch cannot contain future 'Created' & 'Modified' dates
```
eyalbe4: I suggest simplifying the above description, which will hopefully make it clearer.
How about:
// Similar to the SearchResult struct, but without the date fields.
// It is used by the tests, which cannot predict the date fields values.

**(Notice also the added white-space after //)**

utils/tests/searchresultnodate.go
Postyy: Removed 'SearchCommandNoDate' struct.
Created 'SearchResultNoDate' method instead.

utils/tests/searchresultnodate.go
Postyy: Removed.

artifactory_test.go
```diff
@@ -3142,12 +3145,36 @@ func TestArtifactorySearchIncludeDir(t *testing.T) {
...
 	searchCmd.SetSpec(searchSpecBuilder.IncludeDirs(true).BuildSpec())
 	assert.NoError(t, searchCmd.Search())
-	assert.ElementsMatch(t, searchCmd.SearchResult(), tests.GetSearchIncludeDirsFiles())
+	assert.ElementsMatch(t, searchCmd.SearchResultNoDate(), tests.GetSearchIncludeDirsFiles())
+
+	// Cleanup
+	cleanArtifactoryTest()
+}
+
+func TestArtifactorySearchDisplayDate(t *testing.T) {
```
eyalbe4: To save execution time, I suggest removing this test, and instead, tale the following two lines from this test:
```
	assert.NotEmpty(t, result[0].Created)
	assert.NotEmpty(t, result[0].Modified)
```
and add them to all the existing search tests functions.

Postyy: Done.

PR #491, opened 2019-11-07
Split build tools tests

Postyy:


buildtools_test.go
```diff
@@ -1,40 +1,17 @@
...
-	"path"
-	"path/filepath"
-	"strconv"
-	"strings"
-	"testing"
 )
 
-const DockerTestImage string = "jfrog_cli_test_image"
-
 func InitBuildToolsTests() {
```
eyalbe4: I'm not sure the *buildtools_test.go* file is needed anymore, because all build tools have their own test files now.

docker_test.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func (image *buildDockerImage) GetEnv() map[string]string {
+	return map[string]string{}
+}
+
+func (image *buildDockerImage) GetStdWriter() io.WriteCloser {
+	return nil
+}
+func (image *buildDockerImage) GetErrWriter() io.WriteCloser {
```
eyalbe4: Missing new-line before function.

docker_test.go
```diff
@@ -0,0 +1,183 @@
...
+	return nil
+}
+func InitDockerTests() {
```
eyalbe4: Missing new-line before function.

go_test.go
```diff
@@ -0,0 +1,550 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+	projectTarget, err = filepath.Abs(projectTarget)
+	if err != nil {
+		t.Error(err)
+	}
+	return projectTarget
+}
+func downloadModFile(specName, wd, subDir string, t *testing.T) []byte {
```
eyalbe4: Missing new-line before function.

gradle_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"testing"
+
+	"github.com/jfrog/jfrog-cli-go/artifactory/commands/gradle"
+	"github.com/jfrog/jfrog-cli-go/artifactory/utils"
+	"github.com/jfrog/jfrog-cli-go/inttestutils"
+	"github.com/jfrog/jfrog-cli-go/utils/tests"
+)
+
+func TestGradleBuildWithServerID(t *testing.T) {
+	initBuildToolsTest(t)
```
eyalbe4: The *initBuildToolsTest* function should probably not exist, because the *buildtools_test.go* file will be removed, unless you think that we need one init function for all build tools. In that case, it this function should probably in the tests utils file.

jcenter_test.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"os"
+	"path"
+	"testing"
+
+	"github.com/jfrog/jfrog-cli-go/artifactory/utils"
+	"github.com/jfrog/jfrog-cli-go/utils/tests"
+)
+
+func TestGetJcenterRemoteDetails(t *testing.T) {
```
eyalbe4: I'm not sure we need to split tests that are related to jcenter to a separate file.

buildtools_test.go
Postyy: You're right, I will merge this file with main_test.go.

jcenter_test.go
Postyy: I will place jcenter's tests in artifactory's test.

docker_test.go
```diff
@@ -0,0 +1,183 @@
...
+func (image *buildDockerImage) GetStdWriter() io.WriteCloser {
+	return nil
+}
+
+func (image *buildDockerImage) GetErrWriter() io.WriteCloser {
+	return nil
+}
+
+func InitDockerTests() {
+	if !*tests.TestDocker {
```
eyalbe4: This if statement is redundant, because this condition is already checked in the calling function.

main_test.go
```diff
@@ -19,15 +34,14 @@ func TestMain(m *testing.M) {
 
 func setupIntegrationTests() {
 	flag.Parse()
-	*tests.RtUrl = utils.AddTrailingSlashIfNeeded(*tests.RtUrl)
 
 	if *tests.TestBintray {
 		InitBintrayTests()
```
eyalbe4: Inside the *InitBintrayTests()* function, the first if statement is redundant.

main_test.go
```diff
@@ -19,15 +34,14 @@ func TestMain(m *testing.M) {
...
 	}
 	if *tests.TestArtifactory && !*tests.TestArtifactoryProxy {
 		InitArtifactoryTests()
```
eyalbe4: Inside the *InitArtifactoryTests* function, the first if statement has now become redundant.

docker_test.go
```diff
@@ -0,0 +1,183 @@
...
+		return
+	}
+	os.Setenv(cliutils.ReportUsage, "false")
+	os.Setenv(cliutils.OfferConfig, "false")
```
eyalbe4: Setting the above two environment variables can be placed in the beginning of the *setupIntegrationTests()* function. That will allow us to remove the *InitDockerTests()* function, as well as the other init functions, which include only these two env vars set.

maven_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func createMavenProject(t *testing.T) string {
+	srcPomFile := filepath.Join(filepath.FromSlash(tests.GetTestResourcesPath()), "mavenproject", "pom.xml")
+	pomPath, err := tests.ReplaceTemplateVariables(srcPomFile, "")
+	if err != nil {
+		t.Error(err)
+	}
+	return pomPath
+}
+func initMavenTest(t *testing.T) {
```
eyalbe4: Missing new-line before function.

nuget_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// This is due to Artifactory bug, we cant create remote repository with REST API.
+	if !isRepoExist(tests.NugetRemoteRepo) {
+		t.Error("Create nuget remote repository:", tests.NugetRemoteRepo, "in order to run nuget tests")
+		t.FailNow()
+	}
+}
+
+func TestNugetResolve(t *testing.T) {
+	initNugetTest(t)
```
eyalbe4: It looks like the *initNugetTest* function is not invoked from where the other init functions are invoked.

pip_test.go
```diff
@@ -0,0 +1,380 @@
...
+)
+
+type PipCmd struct {
+	Command string
+	Options []string
+}
+
+func TestPipInstall(t *testing.T) {
+	// Init pip.
+	initPipTest(t)
```
eyalbe4: It looks like the *initPipTest* function is not invoked from where the other init functions are invoked.

PR #501, opened 2019-11-21
Fix timestamp proprty for mvn/gradle

Postyy:


artifactory/utils/buildinfoproperties.go
```diff
@@ -202,6 +204,11 @@ func CreateBuildInfoPropertiesFile(buildName, buildNumber string, config *viper.
 		return "", err
 	}
 
+	err = SaveBuildTimestamp(buildName, buildNumber, config)
```
eyalbe4: Since the *SaveBuildTimestamp* function is used only from this file, it may be better to move it into this file and male it private.
Also, to make the code more consistent, let's call it right after the calls to the *setServerDetailsToConfig* and *setServerDetailsToConfig* functions, and rename it to *setBuildTimestampToConfig*.

Postyy: Done.

PR #513, opened 2019-12-02
Bug fix for nuget bp

Or-Gabay:


artifactory/utils/nuget/solution/solution.go
```diff
@@ -50,16 +50,20 @@ func (solution *solution) BuildInfo(module string) (*buildinfo.BuildInfo, error)
...
-		}
-		module := buildinfo.Module{Id: module, Dependencies: projectDependencies}
+		module := buildinfo.Module{Id: getModuleId(module, project.Name()), Dependencies: projectDependencies}
 		modules = append(modules, module)
 	}
 	buildInfo.Modules = modules
 	return buildInfo, nil
 }
 
+func getModuleId(byFlag, byProject string) string {
```
eyalbe4: I suggest renaming the *byFlag* and *byProject* arguments to *customModuleID* and *projectName* respectively.

Or-Gabay: Done.

Issue #511, opened 2019-11-29
CLI claims it couldn't find plexus-classworlds-x.x.x.jar with maven 3.6.3


TimDG:
Hi,

I'm trying to get jfrog cli to run inside a docker image. However it always complains that the plexus classworlds jar is not found.

I've made sure that the M2_HOME env variable points to the right location and I can see that the jar is actually present in the boot folder in there:

```
root@0e0533fc62b3:/source# ./jfrog rt mvn "clean package" .jfrog/mvnconfig.yml 
[Info] Running Mvn...
[Info] Downloading bintray/jcenter/org/jfrog/buildinfo/build-info-extractor-maven3/2.13.13/build-info-extractor-maven3-2.13.13-uber.jar
[Info] Downloaded 1 artifact.
[Error] couldn't find plexus-classworlds-x.x.x.jar in Maven installation path, please check M2_HOME environment variable
root@0e0533fc62b3:/source# ls -l $M2_HOME/boot/
total 72
drwxr-xr-x 2 root root  4096 Nov 28 17:26 .
drwxr-xr-x 1 root root  4096 Nov 28 17:26 ..
-rw-r--r-- 1 root root 52873 Nov  7 12:32 plexus-classworlds-2.6.0.jar
-rw-r--r-- 1 root root 11560 Nov  7 12:32 plexus-classworlds.license
root@0e0533fc62b3:/source# 
```

I've used this docker image: maven:3.6.3-jdk-8 
With maven:3.6.2-jdk-8 it works as intended

TimDG:
To be a bit more complete: I installed jfrog using the `curl -fL https://getcli.jfrog.io | sh` script

ingokofler:
Also affected by this one. Issue is related to latest DockerHub image and the way how they implemented the check...

https://github.com/jfrog/jfrog-cli/blob/master/artifactory/commands/mvn/mvn.go#L153-L163

```go
plexusClassworlds, err := filepath.Glob(filepath.Join(mavenHome, "boot", "plexus-classworlds*"))
```

will find both the .jar and the .license file. The .license file does not exist in previous Maven Docker image versions. So the plexusClassworlds will have two elements and the following check then fails

```go
if len(plexusClassworlds) != 1 {
		return nil, errorutils.CheckError(errors.New("couldn't find plexus-classworlds-x.x.x.jar in Maven installation path, please check M2_HOME environment variable"))
	}
```

[...]

PR #514, opened 2019-12-02
Fix for - CLI claims it couldn't find plexus-classworlds-x.x.x.jar

eyalbe4:
This fixes - https://github.com/jfrog/jfrog-cli/issues/511


shapeofarchitect:
which Docker Image this fix went in ? How Do I check that . Please help. !

PR #523, opened 2019-12-12
Docker commands - Skip missing foreign layers when collecting build-info

barbelity:


artifactory/utils/docker/buildinfo.go
```diff
@@ -200,7 +209,17 @@ func (builder *buildInfoBuilder) handlePush(manifestArtifact, configLayerArtifac
 	return nil
 }
 
-// Set build properties on docker image layers in Artifactory
+func (builder *buildInfoBuilder) handleMissingLayer(layerMediaType, layerFileName string) error {
+	// Allow missing layer to be of a foreign type.
+	if layerMediaType == ForeignLayerMediaType {
+		log.Info(fmt.Sprintf("Foreign layer: %s is missing from Artifactory, thus will not be added to the build-info.", layerFileName))
```
eyalbe4: ...is missing from Artifactory, thus will not be added... --> ...is missing in Artifactory and therefore will not be added...

PR #526, opened 2019-12-14
Make Delete multithreaded

AlexeiVainshtein:


artifactory/commands/generic/delete.go
```diff
@@ -82,11 +91,12 @@ func (dc *DeleteCommand) GetPathsToDelete() error {
 }
 
 func (dc *DeleteCommand) DeleteFiles() (successCount, failedCount int, err error) {
+	var progressBar ioUtils.Progress
```
eyalbe4: The current progress bar implementation measures the progress of an action on a single file. It is probably better therefore not to use progress bar of the delete operation.

PR #535, opened 2019-12-19
Artifactory download - check sync-deletes path does exists before running.

liron-shalom:


artifactory/commands/generic/download.go
```diff
@@ -111,10 +112,18 @@ func (dc *DownloadCommand) Run() error {
 		dc.result.SetFailCount(0)
 		return err
 	} else if dc.SyncDeletesPath() != "" {
-		walkFn := createSyncDeletesWalkFunction(filesInfo)
-		err = fileutils.Walk(dc.SyncDeletesPath(), walkFn, false)
+		absSyncDeletesPath, err := filepath.Abs(dc.SyncDeletesPath())
 		if err != nil {
-			return err
+			return nil
```
eyalbe4: Any reason why return nul here?
We should also have CheckError on the returned err.

artifactory/commands/generic/download.go
```diff
@@ -111,10 +112,18 @@ func (dc *DownloadCommand) Run() error {
...
+		}
+		if _, err = os.Stat(absSyncDeletesPath); err == nil {
+			walkFn := createSyncDeletesWalkFunction(filesInfo)
+			err = fileutils.Walk(dc.SyncDeletesPath(), walkFn, false)
+			if err != nil {
+				return err
```
eyalbe4: We should have CheckError here.

artifactory/commands/generic/download.go
```diff
@@ -111,10 +112,18 @@ func (dc *DownloadCommand) Run() error {
...
+			}
+		} else if os.IsNotExist(err) {
+			log.Info("Sync-deletes path does not exists.")
```
eyalbe4: I suggest adding the value of *absSyncDeletesPath* to this info log message.

PR #543, opened 2019-12-23
Upload cmd - add test

Or-Gabay:


artifactory_test.go
```diff
@@ -4037,3 +4038,20 @@ func initVcsTestDir(t *testing.T) string {
 	}
 	return path
 }
+func searchForDuplicatedVcsProps(props []rtutils.Property, t *testing.T) {
```
eyalbe4: @Or-Gabay,
This function is probably redundant.
You can easily check for duplicates in the *TestVcsProps* function.

Or-Gabay: Fixed here: https://github.com/jfrog/jfrog-cli/pull/548

PR #546, opened 2019-12-23
Gradle & Maven - run tests with cli

Or-Gabay:


utils/tests/utils.go
```diff
@@ -237,6 +237,25 @@ func (cli *JfrogCli) Exec(args ...string) error {
...
+		output = append(output, v)
+	}
+	if cli.credentials != "" {
+		args := strings.Split(cli.credentials, spaceSplit)
+		os.Args = append(os.Args, args...)
+	}
+
+	log.Info("[Command]", strings.Join(output, " "))
+	return cli.main()
+}
```
yahavi: This code is almost identical to the above function `Exec`. Can you consolidate them?

utils/tests/utils.go
```diff
@@ -237,6 +237,25 @@ func (cli *JfrogCli) Exec(args ...string) error {
 	log.Info("[Command]", strings.Join(output, " "))
 	return cli.main()
 }
```
yahavi:
```suggestion
 }

```

main_test.go
```diff
@@ -42,6 +45,7 @@ func setupIntegrationTests() {
 	if *tests.TestNpm || *tests.TestGradle || *tests.TestMaven || *tests.TestGo || *tests.TestNuget || *tests.TestPip {
 		if artifactoryCli == nil {
 			initArtifactoryCli()
+			buildtoolsCreds = authenticate()
```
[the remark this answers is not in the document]
Or-Gabay: Removed.

utils/tests/utils.go
Or-Gabay: Consolidate is a bit overkill since we will delete this functionality when cli 2.0 will be released.

PR #557, opened 2020-01-12
Buildtools - Fix config api-key error message 

Or-Gabay:


artifactory/utils/buildinfoproperties.go
```diff
@@ -259,7 +259,7 @@ func setServerDetailsToConfig(contextPrefix string, vConfig *viper.Viper) error
 	vConfig.Set(contextPrefix+URL, artDetails.GetUrl())
 
 	if artDetails.GetApiKey() != "" {
-		return errorutils.CheckError(errors.New("Server ID " + serverId + " API key authentication is not supported"))
+		return errorutils.CheckError(errors.New("Server ID " + serverId + " must contain user with API key authentication"))
```
eyalbe4: Not sure the new message is clear. I suggest changing it to:
**Configuring an API key without a username is not supported.**
Also notice that the Server ID prefix also makes the sentence unclear. I would just remove it.
The same goes for the URL validation a few lines above.

PR #559, opened 2020-01-13
Issues/550 + fix Gradle config

Or-Gabay:


artifactory/commands/utils/configfile.go
```diff
@@ -125,6 +129,31 @@ func goConfigeFile(configResult *ConfigFile) error {
 	return npmConfigeFile(configResult)
 }
 
+func readGradleGlobalConfig(configResult *ConfigFile) error {
```
eyalbe4: The code in this function seems to be duplicated in the code.
I suggest we stop supporting the old *mvnc* and *gradlec* commands format. Since these commands are interactive (can be run by humans only), we can safely stop supporting them.

artifactory/utils/buildinfoproperties.go
```diff
@@ -58,6 +58,8 @@ const MAVEN_DESCRIPTOR = "deployMavenDescriptors"
 const IVY_DESCRIPTOR = "deployIvyDescriptors"
 const IVY_PATTERN = "ivyPattern"
 const ARTIFACT_PATTERN = "artifactPattern"
+const USE_GRADLE_PLUGIN = "usePlugin"
```
eyalbe4: usePlugin --> usesPlugin

artifactory/utils/buildinfoproperties.go
```diff
@@ -58,6 +58,8 @@ const MAVEN_DESCRIPTOR = "deployMavenDescriptors"
...
+const USE_GRADLE_WRAPPER = "useWrapper"
```
eyalbe4: useWrapper --> usesWrapper

artifactory/commands/utils/configfile.go
Or-Gabay: Removed.

PR #560, opened 2020-01-14
Allow 'go-generate' of completion scripts

yahavi:
Allow generation of bash and zsh completion scripts by running `go generate ./completion/shells/...` from the root dir.
This will allow auto installation of autocomplete scripts by Homebrew:

```ruby
system "go", "generate", "./completion/shells/..."
bash_completion.install "completion/shells/bash/jfrog"
zsh_completion.install "completion/shells/zsh/jfrog" => "_jfrog"
```

### DIFF
--- a/completion/shells/generate_scripts_test.go
+++ b/completion/shells/generate_scripts_test.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"github.com/jfrog/jfrog-cli-go/completion/shells/bash"
+	"github.com/jfrog/jfrog-cli-go/completion/shells/zsh"
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+)
+
+func TestGenrateScripts(t *testing.T) {
+	bashPath := filepath.Join("bash", "jfrog")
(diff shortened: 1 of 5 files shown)


completion/shells/generate_scripts_test.go
```diff
@@ -0,0 +1,42 @@
...
+	// Run go generate ./...
+	cmd := exec.Command("go", "generate", "./...")
+	err := cmd.Run()
+	assert.NoError(t, err)
+
+	// Check bash completion script
+	file, err := os.Open(bashPath)
+	assert.NoError(t, err)
+	b, err := ioutil.ReadAll(file)
+	file.Close()
```
eyalbe4: Let's do this with defer right after the file is created.

completion/shells/generate_scripts_test.go
```diff
@@ -0,0 +1,42 @@
...
+	assert.Equal(t, bash.BashAutocomplete, string(b))
+
+	// Check zsh completion script
+	file, err = os.Open(zshPath)
+	assert.NoError(t, err)
+	b, err = ioutil.ReadAll(file)
+	file.Close()
```
eyalbe4: Let's do this with defer right after the file is created.

PR #561, opened 2020-01-14
Docker - validate docker version

Or-Gabay:


artifactory/utils/docker/docker.go
```diff
@@ -314,3 +322,43 @@ func DockerLogin(imageTag string, config *DockerLoginConfig) error {
...
+
+func (versionCmd *versionCmd) GetStdWriter() io.WriteCloser {
+	return nil
+}
+
+func (versionCmd *versionCmd) GetErrWriter() io.WriteCloser {
+	return nil
+}
+
+func IsVersionSupported() error {
```
eyalbe4: Functions with the *is* prefix usually return a boolean.
Therefore, either have this function return a boolean, or rename it to *validateVersion*.
I think that it is better to return a boolean, so that errors that is caused due to execution failure are distinguishable than unsupported versions.

artifactory/commands/docker/pull.go
```diff
@@ -23,6 +24,12 @@ func (dpc *DockerPullCommand) Run() error {
 	if errorutils.CheckError(err) != nil {
 		return err
 	}
+
+	err = docker.IsVersionSupported()
```
eyalbe4: Any reason to do *docker login* with unsupported docker versions?
The login in that case will fail...

Or-Gabay: Moved to the beginning of the function

artifactory/utils/docker/docker.go
Or-Gabay: Done

PR #570, opened 2020-01-17
 Implement token commands

SyntacticStructures:
* Get Tokens
* Create Token
* Refresh Token
* Revoke Token

Add tests for auth token commands


docs/artifactory/refreshtoken/help.go
```diff
@@ -0,0 +1,12 @@
+package refreshtoken
+
+const Description = "Refresh access token."
+
+var Usage = []string{"jfrog rt refresh-token"}
+
+const Arguments string = `	refresh token
+		The refresh token of the access token that needs to be refreshed.
```
eyalbe4: The refresh token of the access token that needs to be refreshed. --> The refresh token used for refreshing the access token.

artifactory/commands/token/revoketoken.go
```diff
@@ -0,0 +1,41 @@
+package token
+
+import (
+	"github.com/jfrog/jfrog-cli-go/artifactory/utils"
+	"github.com/jfrog/jfrog-client-go/artifactory/services"
+)
+
+type RevokeTokenCommand struct {
+	TokenCommand
+	params    services.RevokeTokenParams
```
eyalbe4: This comment applies to most of the new token commands:
We try not to use params of types such as *services.RevokeTokenParams* as part of the commands params. This is because *services.RevokeTokenParams* is defined by the jfrog-client-go library, and we don't want to changes the jfrog-client-go API to affect the commands options. See how it is does in other commands as a reference.

artifactory/cli.go
```diff
@@ -1392,6 +1444,44 @@ func getPipInstallFlags() []cli.Flag {
 	return getBuildToolAndModuleFlags()
 }
 
+func getCreateTokenFlags() []cli.Flag {
```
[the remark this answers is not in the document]
SyntacticStructures: I believe they they are used, both in the go client and rest api: https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API#ArtifactoryRESTAPI-RefreshToken

eyalbe4:
@SyntacticStructures,
https://github.com/jfrog/jfrog-cli/pull/667 adds a new ```access-token-create``` command, which is a subset of the content of this pull request.
I'm closing this one for now. Feel free to create a new one if you'd like to continue working in this.

PR #572, opened 2020-01-21
Fix - The "jfrog rt c" command does not accept the client cert flags.

eyalbe4:


artifactory/commands/config.go
```diff
@@ -113,12 +113,6 @@ func (cc *ConfigCommand) Config() error {
 		return err
 	}
 
-	if cc.encPassword {
```
RobiNino: Moving the encryption from here means password provided in interactive=false will not be encrypted

artifactory/commands/config.go
```diff
@@ -193,32 +187,56 @@ func (cc *ConfigCommand) getConfigurationFromUser() error {
 	// Ssh-Key
 	if fileutils.IsSshUrl(cc.details.Url) {
 		return getSshKeyPath(cc.details)
-	} else {
```
RobiNino: Worth to make sure client certificates are supported when using ssh

artifactory/commands/config.go
eyalbe4: You're right. Thanks.

Issue #573, opened 2020-01-21
v1.33.1 help not working for jfrog rt npm* commands


tegataiprime:
We're seeing that for jfrog cli 1.33.1, the --help parameter is not working for several npm commands:
e.g.
```bash
$ jfrog --version
jfrog version 1.33.1

$ jfrog rt npmi --help
[Info] Running npm Install.
[Error] invalid character 'N' looking for beginning of value

$ jfrog rt npmci --help
[Info] Running npm Install.
[Error] invalid character 'N' looking for beginning of value

$ jfrog rt npmp --help
[Info] Running npm Publish
npm pack [[<@scope>/]<pkg>...] [--dry-run]
[Error] Two errors occurred: 
open /opt/source/myapp-0.0.0.tgz: no such file or directory 
remove /opt/source/myapp-0.0.0.tgz: no such file or directory
```

`jfrog rt npmc --help` does work.

These commands work when using version 1.32.4.

eyalbe4:
Thank you for reporting this @tegataiprime
https://github.com/jfrog/jfrog-cli/pull/575 fixes this issue.

tegataiprime:
Sweet, thank you :)

eyalbe4:
@tegataiprime - this has been released already.
Feel free to close this issue.
Thanks.

PR #575, opened 2020-01-23
The --help option is not working for some commands.

eyalbe4:
Fixes https://github.com/jfrog/jfrog-cli/issues/573


Or-Gabay:
We already have functions that skip parsing flags for each buildtool.
For example, we can add `os.Args[3] == "--help"` to `shouldSkipNpmFlagParsing` so code-gangsta will catch the --help and use it to show  npm help.

PR #583, opened 2020-01-27
Improve Artifactory Tests

RobiNino:


artifactory_test.go
```diff
@@ -423,10 +386,8 @@ func TestArtifactoryUploadFilesNameWithParenthesis(t *testing.T) {
```
eyalbe4: Missing errro assertion in this function.

PR #585, opened 2020-01-29
Docker - fix script for get docker version on windows

Or-Gabay:


artifactory/utils/docker/docker_test.go
```diff
@@ -89,3 +94,12 @@ func TestCheckDockerMinVersion(t *testing.T) {
 		t.Errorf("checkDockerMinVersion(%s) == %t, want %t", have, got, want)
 	}
 }
+
+func TestDockerVersionScript(t *testing.T) {
+	cmd := &versionCmd{}
+	content, err := gofrogcmd.RunCmdOutput(cmd)
+	if err != nil || strings.Index(content, "Docker version") == -1 {
+		t.Error("Could not get docker version")
```
eyalbe4: If err is not nil, let's print its message as well.

artifactory/utils/docker/docker_test.go
```diff
@@ -89,3 +94,12 @@ func TestCheckDockerMinVersion(t *testing.T) {
...
+	}
+
```
eyalbe4: Redundant new-line.

PR #588, opened 2020-01-30
Add 'target' flag for Xray offline-update command

barbelity:


xray/cli.go
```diff
@@ -45,13 +45,18 @@ func offlineUpdateFlags() []cli.Flag {
 			Name:  "version",
 			Usage: "[Optional] Xray API version.` `",
 		},
+		cli.StringFlag{
+			Name:  "target",
+			Usage: "[Optional] Path for downloaded update files.` `",
```
eyalbe4: [Optional] --> [Default: ./]

Question - Did we check that the target option works both with and without a trailing slash? Did test this these two options on Windows as well?

barbelity: Checked.

PR #602, opened 2020-02-20
Tests - Delete Go and and Pypi repositories

yahavi:


artifactory_test.go
```diff
@@ -3354,6 +3354,7 @@ func createRandomReposName() {
 	}
 }
 
+// Important - Virtual repositories most be deleted first
 func deleteRepos() {
 	repos := []string{
 		tests.VirtualRepo,
```
eyalbe4: It looks like the following repositories are deleted (and also created) regardless of the tests suite we run. For example, they will be created and deleted for maven tests, even if they are not used there.
I guess that we can save some tests time if we created them only in tests suite where they are needed.
```
	repos := []string{
		tests.VirtualRepo,
		tests.Repo1,
		tests.Repo2,
		tests.LfsRepo,
		tests.DebianRepo,
		tests.JcenterRemoteRepo,
	}
```

yahavi: We agreed to implement this in the near future.

PR #607, opened 2020-02-25
Disable all interactive prompts when CI=true

eyalbe4:


utils/cliutils/utils.go
```diff
@@ -137,6 +139,31 @@ func PrintHelpAndReturnError(msg string, context *cli.Context) error {
...
 
+func GetQuietValue(c *cli.Context) bool {
+	if c.IsSet("quiet") {
+		return c.Bool("quiet")
+	}
+
+	return getCiValue()
+}
+
+func GetInteractiveValue(c *cli.Context) bool {
```
liron-shalom: I thing this logic worth some comment to explain the connection between "CI" and "interactive"

main_test.go
```diff
@@ -27,9 +27,7 @@ func TestMain(m *testing.M) {
 
 func setupIntegrationTests() {
 	os.Setenv(cliutils.ReportUsage, "false")
-	os.Setenv(cliutils.OfferConfig, "false")
-	// Disable progress bar:
```
liron-shalom: Let's keep the comment and add :
// Disable progress bar and confirmation messages

utils/cliutils/utils.go
```diff
@@ -137,6 +139,31 @@ func PrintHelpAndReturnError(msg string, context *cli.Context) error {
 	return errors.New(msg)
 }
 
+func GetQuietValue(c *cli.Context) bool {
```
liron-shalom: I think this logic worth some comment to explain the connection between "CI" and "quiet"

PR #624, opened 2020-03-10
Docker - support fat manifest

Or-Gabay:


artifactory/utils/docker/buildinfo.go
```diff
@@ -316,7 +317,44 @@ func getConfigLayer(imageId string, searchResults map[string]utils.ResultItem, s
 }
 
 // Search for image layers in Artifactory.
-func searchImageLayers(imageId, imagePathPattern string, serviceManager *artifactory.ArtifactoryServicesManager) (map[string]utils.ResultItem, error) {
+func searchImageLayers(builder *buildInfoBuilder, imagePathPattern string, serviceManager *artifactory.ArtifactoryServicesManager) (map[string]utils.ResultItem, error) {
```
eyalbe4: I suggest adding a unit test, which will checks the parsing of the fat manifest.

docker_test.go
```diff
@@ -181,3 +211,25 @@ func TestDockerVersionScript(t *testing.T) {
 		t.Errorf("Could not get docker version, error: %s", err.Error())
 	}
 }
+
+func TestDockerFatManifestPull(t *testing.T) {
```
eyalbe4: How is this test different from the existing TestDockerPull? Why is it needed?

artifactory/utils/docker/buildinfo.go
```diff
@@ -316,7 +317,44 @@ func getConfigLayer(imageId string, searchResults map[string]utils.ResultItem, s
...
+
+	// Validate image ID layer exists.
+	if _, ok := resultMap[digestToLayer(builder.imageId)]; !ok {
+		// In case of a fat-manifest, Artifactory will create two folders.
+		// One folder named as the image tag, contain inside the fat manifest,
+		// The second folder, named as image's manifest digest, will contain the image layers and the image's manifest.
+		if _, ok := resultMap["list.manifest.json"]; ok {
+			v, _ := builder.image.Manifest()
+			var listManifest []Manifest
+			err := json.Unmarshal([]byte(v), &listManifest)
```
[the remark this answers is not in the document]
Or-Gabay: Add log instead of new error

docker_test.go
Or-Gabay: Docker pull works with a non-fat-manifest image, unlike TestDockerFatManifestPull.
The idea behind this test is to validate the build info while pulling a fat-manifest-image from the remote repo.
The problem with remote repo is that Artifactory makes a different folder structure for pulling fat-manifest-images  witch break the collect build info step

artifactory/utils/docker/buildinfo.go
Or-Gabay: We don't parse but marshal the expected fat-manifest format,
TestDockerFatManifestPull should be all we need to check this

PR #635, opened 2020-03-17
Repo templates - add error handling for template path

liron-shalom:

### DIFF
--- a/artifactory/commands/repository/template.go
+++ b/artifactory/commands/repository/template.go
@@ -477,14 +477,18 @@ func (rtc *RepoTemplateCommand) RtDetails() (*config.ArtifactoryDetails, error) 
 }
 
 func (rtc *RepoTemplateCommand) Run() (err error) {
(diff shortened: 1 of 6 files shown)


liron-shalom:
Which is correct.
At the end we want path to be non-existing file.
So directory is not valid (existing or not) and existing file is not valid as well.

artifactory/commands/repository/template.go
```diff
@@ -473,6 +477,21 @@ func (rtc *RepoTemplateCommand) RtDetails() (*config.ArtifactoryDetails, error)
...
+		return errors.New("Path cannot be a directory." + PathErrorSuffixMsg)
+	}
+	exists, err = fileutils.IsFileExists(rtc.path, true)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return errors.New("File already exists." + PathErrorSuffixMsg)
+	}
+
```
yahavi: Please add errorutils.CheckError(err) to all of the returned errors

PR #642, opened 2020-03-18
Templates - bugfix for unknow tempalt json key

Or-Gabay:


artifactory/commands/replication/create.go
```diff
@@ -68,7 +69,12 @@ func (rcc *ReplicationCreateCommand) Run() (err error) {
 		if key == "serverId" {
 			serverId = value.(string)
 		} else {
-			writersMap[key](&replicationConfigMap, key, value.(string))
+			if writerMapFunc, ok := writersMap[key]; ok {
```
liron-shalom: 1. I think it will be more readable if we will check "!ok", and continue without the "else"
2. The variable name can be shorten just to "writer"

artifactory/commands/repository/repository.go
```diff
@@ -41,7 +42,12 @@ func (rc *RepoCommand) PerformRepoCmd(isUpdate bool) (err error) {
 	// All the values in the template are strings
 	// Go over the the confMap and write the values with the correct type using the writersMap
 	for key, value := range repoConfigMap {
-		writersMap[key](&repoConfigMap, key, value.(string))
+		if writerMapFunc, ok := writersMap[key]; ok {
+			writerMapFunc(&repoConfigMap, key, value.(string))
+		} else {
+			err = errorutils.CheckError(errors.New("Unknown key: \"" + key + "\" for repo create"))
```
liron-shalom: Same for here.

artifactory/commands/replication/create.go
```diff
@@ -68,7 +69,12 @@ func (rcc *ReplicationCreateCommand) Run() (err error) {
...
+				writerMapFunc(&replicationConfigMap, key, value.(string))
+			} else {
+				err = errorutils.CheckError(errors.New("Unknown key: \"" + key + "\" for replication create"))
```
liron-shalom: IntelliJ says that Go error msg should not start with capital letter

artifactory/commands/utils/questionnaire.go
```diff
@@ -276,3 +276,13 @@ func OptionalKeyCallback(iq *InteractiveQuestionnaire, key string) (value string
 	}
 	return value, err
 }
+
+func IsValideParams(key string, value interface{}, writersMap map[string]AnswerWriter) bool {
```
liron-shalom: I suggest rename this function to "validateMapEntry"
and let the return value be an error with the correct msg: "unknown-key" or "non-string-value"

artifactory/commands/utils/questionnaire.go
```diff
@@ -276,3 +278,13 @@ func OptionalKeyCallback(iq *InteractiveQuestionnaire, key string) (value string
 	}
 	return value, err
 }
+
+func ValidateMapEntry(key string, value interface{}, writersMap map[string]AnswerWriter) error {
```
liron-shalom: This function does not belong to the questionnaire module.
We need to find a better place to put it ...

artifactory/commands/replication/create.go
```diff
@@ -68,7 +68,12 @@ func (rcc *ReplicationCreateCommand) Run() (err error) {
 		if key == "serverId" {
 			serverId = value.(string)
 		} else {
-			writersMap[key](&replicationConfigMap, key, value.(string))
+			for key, value := range replicationConfigMap {
```
eyalbe4: This loop iterates over the map, but it is inside an outer loop which also iterates the map.
Any reason why we should iterate the map n squared times?

Or-Gabay: Remove duplicate for loop

artifactory/commands/utils/questionnaire.go
Or-Gabay: Well, you know more than me about this, where should we put it?

PR #647, opened 2020-03-24
Build add git command - add server-id flag

barbelity:


barbelity:
Removed the redundant if statement.

PR #660, opened 2020-03-30
Bugfix - Remove inner quote before parse

Or-Gabay:


artifactory/utils/argsutils_test.go
```diff
@@ -263,4 +263,12 @@ func TestParseArgs(t *testing.T) {
 	want = []string{`--build-name=jfrog jfrog`, "--build-number=1"}
```
eyalbe4: The above *want* is good for the following two tests as well. No need to declare it again.

Or-Gabay: Removed

PR #667, opened 2020-04-02
Add create access token command.

liron-shalom:


artifactory/cli.go
```diff
@@ -1076,6 +1090,31 @@ func getReplicationDeleteFlags() []cli.Flag {
...
+			Name:  "expiry",
+			Usage: "[Default: " + strconv.Itoa(cliutils.TokenExpiry) + "] The time in seconds for which the token will be valid.` `",
+		},
+		cli.BoolFlag{
+			Name:  "refreshable",
+			Usage: "[Default: false] Set to true if you'd like the the token to be refreshable. A refresh token will also be returned in order to be used to generate a new token once it expires.` `",
+		},
+		cli.StringFlag{
+			Name:  "audience",
+			Usage: "[Optional] A space-separate list of the other Artifactory instances or services that should accept this token.` `",
```
eyalbe4, line 1113: I suggest changing this description to the following:
[Optional] A space-separate list of the other Artifactory instances or services that should accept this token identified by their Artifactory Service IDs, as obtained by the 'jfrog rt curl api/system/service_id' command.

docs/artifactory/tokencreate/help.go
```diff
@@ -0,0 +1,8 @@
+package tokencreate
+
+const Description = "Creates an access token. By default an user-scoped token will be created, unless groups and/or admin-privileges are specified."
```
eyalbe4, line 3: ...unless groups and/or admin-privileges are specified. --> ...unless the --groups and/or --grant-admin options are specified.

docs/artifactory/tokencreate/help.go
```diff
@@ -0,0 +1,8 @@
...
+
+var Usage = []string{`jfrog rt tc <user name>`}
```
eyalbe4, line 5: tc --> atc

artifactory_test.go
```diff
@@ -3695,3 +3695,37 @@ func TestArtifactoryReplicationCreate(t *testing.T) {
...
+	// Extract the the token from the output
+	token, err := jsonparser.GetString(content, "access_token")
+	assert.NoError(t, err)
+
+	// Try ping with the new token
+	err = artifactoryCli.Exec("ping", "--access-token="+token)
+	assert.NoError(t, err)
+
+	// Restore previous logger
+	log.SetLogger(previousLog)
```
eyalbe4, line 3728: Let's do this in a defer, right after the logger is replaced.
Since this is done in an additional test as well, let's change it there also.

artifactory/commands/generic/createaccesstoken.go
```diff
@@ -0,0 +1,130 @@
...
+	atcc.groups = groups
+	return atcc
+}
+
+func (atcc *AccessTokenCreateCommand) Response() ([]byte, error) {
+	content, err := json.Marshal(*atcc.response)
+	if errorutils.CheckError(err) != nil {
+		return nil, err
+	}
+	return content, nil
```
eyalbe4, line 72: The above function content can be simplified to:
```
content, err := json.Marshal(*atcc.response)
return content, errorutils.CheckError(err)

```

artifactory/cli.go
```diff
@@ -1076,6 +1090,33 @@ func getReplicationDeleteFlags() []cli.Flag {
...
+				"Specify * to indicate that this is a 'user-scoped token', i.e., the token provides the same access privileges that the current subject has, and is therefore evaluated dynamically. " +
+				"A non-admin user can only provide a scope that is a subset of the groups to which he belongs` `",
+		},
+		cli.BoolFlag{
+			Name:  "grant-admin",
+			Usage: "[Default: false] Set to true to provides admin privileges to the access token. This is only available for administrators.` `",
+		},
+		cli.StringFlag{
+			Name:  "expiry",
+			Usage: "[Default: " + strconv.Itoa(cliutils.TokenExpiry) + "] The time in seconds for which the token will be valid. To specify a token that never expires, set to zero. Non-admin can only set a value that is equal to or less than the default 3600.` `",
```
eyalbe4, line 1107: Is this true?
Non-admin can only set a value that is equal to or less than the default 3600.
Can't a non-admin create an access token for week?

PR #671, opened 2020-04-08
NuGet tests - Remove skipping in case of non-Windows operating system.

liron-shalom:


liron-shalom:
Will open new PR with all non-windows support.

PR #685, opened 2020-05-04
Docker - bugfix duplicate layers in build-info

Or-Gabay:


artifactory/utils/docker/buildinfo.go
```diff
@@ -412,6 +418,19 @@ func (configLayer *configLayer) getNumberOfDependentLayers() int {
 	return layersNum
 }
 
+func FilterDuplicateDockerLayer(imageMLayers []layer) []layer {
```
eyalbe4, line 421: Looks like this function can be made private.
Also, I suggest renaming it to *removeDuplicateDockerLayers*.

artifactory/utils/docker/buildinfo.go
```diff
@@ -277,7 +277,9 @@ func getManifest(imageId string, searchResults map[string]utils.ResultItem, serv
 	if errorutils.CheckError(err) != nil {
 		return nil, buildinfo.Artifact{}, buildinfo.Dependency{}, err
 	}
-
+	// Filter duplicate layers.
```
eyalbe4, line 280: // Filter duplicate layers. --> // Remove duplicate layers.

PR #692, opened 2020-05-12
Make refreshable tokens true by default

RobiNino:


artifactory/commands/config.go
```diff
@@ -238,12 +234,14 @@ func (cc *ConfigCommand) getConfigurationFromUser() error {
 		}
 	}
 
-	cc.readRefreshableTokensFromConsole()
 	cc.readClientCertInfoFromConsole()
```
eyalbe4, line 237: I think that the client cetificates data should be at the very end of the interactive configuration process.

artifactory/cli.go
```diff
@@ -4020,13 +4025,18 @@ func createConfigCommandConfiguration(c *cli.Context) (configCommandConfiguratio
...
 	return
 }
 
 func validateConfigFlags(configCommandConfiguration *commands.ConfigCommandConfiguration) error {
 	if !configCommandConfiguration.Interactive && configCommandConfiguration.ArtDetails.Url == "" {
 		return errors.New("the --url option is mandatory when the --interactive option is set to false or the CI environment variable is set to true.")
 	}
+	// Validate the option is not used along with an access token
+	if configCommandConfiguration.BasicAuthOnly && configCommandConfiguration.ArtDetails.AccessToken != ""{
+		return errors.New("the --basic-auth-only option is only supported with basic authentication options")
```
eyalbe4, line 4038: ...only supported with basic authentication options --> ...only supported when username and password / API key are provided.

PR #726, opened 2020-06-04
Search/Delete/Download/Props - Adjust reader/writer

Or-Gabay:


artifactory/commands/generic/delete.go
```diff
@@ -65,31 +63,45 @@ func (dc *DeleteCommand) Run() error {
...
+		}
+		if length > 0 {
+			temp = append(temp, currentResultsReader)
+		} else {
+			currentResultsReader.Close()
 		}
-		dc.deleteItems = append(dc.deleteItems, currentResultItems...)
 	}
-	return nil
+	defer func() {
```
eyalbe4, line 96: It is better to add the close of the readers with "ddefer" right after each reader is returned. There's probably no need for this loop. The way it is done now, if an error is returned before the function ends, these readers can be left open.
Also, there's another currentResultsReader.Close() above, in case the length is 0. I'm not sure we need to handle this case seperately (and without a defer).

artifactory/commands/generic/deleteprops.go
```diff
@@ -31,15 +31,19 @@ func (deleteProps *DeletePropsCommand) Run() error {
...
-	resultItems, searchErr := searchItems(deleteProps.Spec(), servicesManager)
-
-	propsParams := GetPropsParams(resultItems, deleteProps.props)
+	reader, searchErr := searchItems(deleteProps.Spec(), servicesManager)
+	if searchErr != nil {
+		return searchErr
+	}
+	defer reader.Close()
+	propsParams := GetPropsParams(reader, deleteProps.props)
 	success, err := servicesManager.DeleteProps(propsParams)
```
eyalbe4, line 40: This error is unchecked.

artifactory/commands/generic/props.go
```diff
@@ -60,33 +62,45 @@ func createPropsServiceManager(threads int, artDetails *config.ArtifactoryDetail
...
 		}
-		resultItems = append(resultItems, currentResultItems...)
+		for searchResult := new(clientutils.ResultItem); reader.NextRecord(searchResult) == nil; searchResult = new(clientutils.ResultItem) {
+			writer.Write(*searchResult)
+		}
+		if err = reader.GetError(); err != nil {
+			errorOccurred = true
+			log.Error(err)
+		}
+		reader.Close()
```
eyalbe4, line 92: Let's close this with defer above (right after it is received).

artifactory/commands/generic/deleteprops.go
Or-Gabay: We return it eventually but not right away, we want to fill the Results() for the print summary.

artifactory/commands/generic/search.go
```diff
@@ -22,93 +26,125 @@ type SearchResult struct {
...
 		}
-		resultItems = append(resultItems, currentResultItems...)
+		searchResults = append(searchResults, currentReader)
 	}
-
-	sc.searchResult = aqlResultToSearchResult(resultItems)
-	clientutils.LogSearchResults(len(resultItems))
-	return err
+	defer func() {
+		for _, reader := range searchResults {
```
eyalbe4, line 97: Similar to a previous comment I added here, it is best to add the defer right after *currentReader* is received, and not in a loop here.

artifactory/commands/generic/search.go
```diff
@@ -22,93 +26,125 @@ type SearchResult struct {
...
+			reader.Close()
+		}
+	}()
+	reader, err := aqlResultToSearchResult(searchResults)
+	if err != nil {
+		return nil, err
+	}
+	length, err := reader.Length()
+	clientartutils.LogSearchResults(length)
+	return reader, err
```
eyalbe4, line 107: Do we need to reset the reader before returning it, or does Length() take care of it already?

artifactory/commands/generic/search.go
```diff
@@ -122,3 +158,36 @@ func GetSearchParams(f *spec.File) (searchParams services.SearchParams, err erro
 	searchParams.IncludeDirs, err = f.IsIncludeDirs(false)
 	return
 }
+
+func PrintSearchResults(reader *content.ContentReader) error {
```
eyalbe4, line 162: The logic of this function can probably be simplified so that it is easier to read and understand.
Let's also add a unit test for this function.

artifactory/commands/generic/search.go
```diff
@@ -122,3 +158,36 @@ func GetSearchParams(f *spec.File) (searchParams services.SearchParams, err erro
...
+	}
+	if prevSearchResult != nil {
+		performPrintSearchResults(*prevSearchResult, "")
+	}
+	log.Output("]")
+	reader.Reset()
+	return reader.GetError()
+}
+
+func performPrintSearchResults(toPrint SearchResult, suffix string) error {
```
eyalbe4, line 186: performPrintSearchResults --> printSearchResult

artifactory/commands/generic/setprops.go
```diff
@@ -31,16 +31,20 @@ func (setProps *SetPropsCommand) Run() error {
...
-	propsParams := GetPropsParams(resultItems, setProps.props)
+	reader, searchErr := searchItems(setProps.Spec(), servicesManager)
+	defer reader.Close()
+	propsParams := GetPropsParams(reader, setProps.props)
 	success, err := servicesManager.SetProps(propsParams)
 
 	result := setProps.Result()
 	result.SetSuccessCount(success)
-	result.SetFailCount(len(resultItems) - success)
+	totalLength, totalLengthErr := reader.Length()
```
eyalbe4, line 41: This error is not checked.

artifactory/commands/generic/setprops.go
```diff
@@ -31,16 +31,20 @@ func (setProps *SetPropsCommand) Run() error {
...
+	result.SetFailCount(totalLength - success)
 	if err == nil {
```
eyalbe4, line 43: This error is checked too late.

artifactory/commands/generic/search.go
Or-Gabay: The `Length()` take care of it.

artifactory/utils/docker/buildinfo.go
```diff
@@ -384,15 +396,17 @@ func performSearch(imagePathPattern string, serviceManager *artifactory.Artifact
...
 	}
 	resultMap := map[string]utils.ResultItem{}
-	for _, v := range results {
-		resultMap[v.Name] = v
+	for resultItem := new(utils.ResultItem); reader.NextRecord(resultItem) == nil; resultItem = new(utils.ResultItem) {
+		resultMap[resultItem.Name] = *resultItem
+
 	}
-	return resultMap, nil
+	defer reader.Close()
```
eyalbe4, line 408: Let's move this Close() up (right after the error check).

artifactory/utils/delete.go
```diff
@@ -2,20 +2,30 @@ package utils
...
+	"github.com/jfrog/jfrog-client-go/utils/io/content"
 )
 
-func ConfirmDelete(pathsToDelete []rtclientutils.ResultItem) bool {
-	if len(pathsToDelete) < 1 {
-		return false
+func ConfirmDelete(pathsToDeleteReader *content.ContentReader) (bool, error) {
+	length, err := pathsToDeleteReader.Length()
+	if err != nil {
+		return false, nil
```
eyalbe4, line 16: We're retuning nil if the error is not nil. Looks like a bug.

artifactory/utils/delete.go
```diff
@@ -2,20 +2,30 @@ package utils
...
+	}
+	if length < 1 {
+		return false, nil
+	}
```
eyalbe4, line 20: Looks like the above two if statements can be merged into one:
```
if err != nil || length < 1 {
    return false, err
}
```

testsdata/prefix/a
```diff
@@ -0,0 +1 @@
+a/a1.in
```
eyalbe4, line 1: This PR adds some files under the testsdata directory. I think @RobiNino renamed the testsdata directory in a different PR, which hasn't been merged yet, so I suggest you sync with him to avoid recreating the directory he renamed.

artifactory_test.go
```diff
@@ -763,16 +763,17 @@ func TestArtifactorySelfSignedCert(t *testing.T) {
 	// Without loading the certificates (or skipping verification) we expect all actions to fail due to error: "x509: certificate signed by unknown authority"
 	searchCmd := generic.NewSearchCommand()
 	searchCmd.SetRtDetails(artifactoryDetails).SetSpec(fileSpec)
-	err = searchCmd.Search()
+	_, err = searchCmd.Search()
```
eyalbe4, line 766: The reader returned by this function should be closed.

artifactory_test.go
```diff
@@ -820,7 +822,10 @@ func TestArtifactoryClientCert(t *testing.T) {
 	// Without loading a valid client certificate, we expect all actions to fail due to error: "tls: bad certificate"
 	searchCmd := generic.NewSearchCommand()
 	searchCmd.SetRtDetails(artifactoryDetails).SetSpec(fileSpec)
-	err = searchCmd.Search()
+	reader, err := searchCmd.Search()
+	if reader != nil {
```
eyalbe4, line 826: The reader returned by the ```searchCmd.Search()``` function can be nil? If so, why aren't we doing nil checks it in order places before closing it? (see other usages in this file, and possible in other places).

artifactory_test.go
```diff
@@ -1756,16 +1770,48 @@ func TestArtifactoryDeleteExclusionsBySpec(t *testing.T) {
 	cleanArtifactoryTest()
 }
 
+// Deleting files which one file name is a prefix to another in the same dir
```
eyalbe4, line 1773: which --> when

artifactory_test.go
Or-Gabay: The test supposed to fail and validate the output error, if something goes wrong, we should close the reader and fail for error validation.

Issue #728, opened 2020-06-06
Gradle Proxy Enablement ignores no_proxy rules


Enquier:
**Describe the bug**
The changes introduced in #359 while important disable the ability to specify a proxy for resolving external dependencies while publishing to an internal repository. For example you might be pulling from an external repository such as external jcenter() or in my case a non-ivy non-maven source that needs manipulation before it can be added to artifactory

**To Reproduce**

1. Configure jfrog/gradle to use http/https_proxy with no_proxy for internal sites (ive tried both IP's and DNS names)
1. Configure a dependency resolution repository that requires a proxy to access
1. Perform gradle tasks including dependency resolution via proxy
1.  Publish Artifacts to internal Artifactory

**Expected behavior**
Gradle/Artifactory plugin to respect no_proxy rules

**Versions**
- JFrog CLI version: Latest
- JFrog CLI operating system: debian/docker
- Artifactory Version: 6.17.0

**Additional context**
Gradle: 4.10.2

eyalbe4:
@Enquier,
Thanks for reporting this. We should probability add a "bypass http proxy" option when configuring an Artifactory server through the *jfrog rt c* command.  What do you think?

Enquier:
That would be a great way to manage internal vs external rt servers/repositories

Enquier:
I have added a PR with a fix for this. CLA is not available via link anymore

joanna-polec:
Hi, do you have any update about this issue?
I have similar problem with gradle deployment, it ignores NO_PROXY env (there is no such issue for mvn or npm). Only one workaround which works is to unset HTTP_PROXY env before running  `jf gradle artifactoryPublish...`
jfrog cli version: 2.59.1

yahavi:
@joanna-polec  
JFrog CLI version 2.65.0 has been released, and it includes a fix for the no-proxy issue during deployment. Please give it a try and let us know your feedback.

PR #729, opened 2020-06-07
Adding fix for and tests for no proxy support for deployer

Enquier:
- [ ] I signed [JFrog's CLA](https://secure.echosign.com/public/hostedForm?formid=5IYKLZ2RXB543N). This is a dead link btw

- [X ] All [tests](https://github.com/jfrog/jfrog-cli#tests) passed. If this feature is not already covered by the tests, I added new tests.
- [X ] This pull request is on the dev branch.
- [ X] I used gofmt for formatting the code before submitting the pull request.
-----
This is a quick fix for #728 . I used the resolveProxyUrl from http to check if the deployer repo needs the proxy or not. Tests are added. There is no setting for resolver proxy in the artifactory gradle plugin. The proxy for resolver is handled by gradle/system so we only have control over the deployer proxy opts in this state. Previously this would only look at the proxy env settings and just blindly set the proxy

### DIFF
--- a/artifactory/utils/buildinfoproperties.go
+++ b/artifactory/utils/buildinfoproperties.go
@@ -2,8 +2,10 @@ package utils
 
 import (
 	"errors"
+	"github.com/jfrog/jfrog-client-go/utils/log"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -20,6 +22,7 @@ import (
 
 const (
 	HttpProxy = "HTTP_PROXY"
+	NoProxy   = "NO_PROXY"
 )
 
 type BuildConfigMapping map[ProjectType][]*map[string]string
@@ -235,6 +238,9 @@ func CreateBuildInfoPropertiesFile(buildName, buildNumber string, config *viper.
 func setProxyIfDefined(config *viper.Viper) error {
 	// Add HTTP_PROXY if exists
 	proxy := os.Getenv(HttpProxy)
+	noProxy := os.Getenv(NoProxy)
+	log.Debug("Proxy:" + proxy)
+	log.Debug("No Proxy:" + noProxy)
 	if proxy != "" {
 		url, err := url.Parse(proxy)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


artifactory/utils/buildinfoproperties.go
```diff
@@ -244,8 +250,29 @@ func setProxyIfDefined(config *viper.Viper) error {
...
+			deployRequest, err := http.NewRequest("HEAD", config.GetString(DEPLOYER_PREFIX+URL), nil)
+			if err != nil {
+				return errorutils.CheckError(err)
+			}
+			deployProxyURL, err := http.ProxyFromEnvironment(deployRequest)
+			log.Debug("Deploy URL:" + config.GetString(DEPLOYER_PREFIX+URL))
+			if err != nil {
+				return errorutils.CheckError(err)
+			}
+			if deployProxyURL != nil {
```
eyalbe4, line 264: There's a chance I'm missing something here, but it looks like ```deployProxyURL``` is used only for the debug logging. My guess is that your intention was to extract the host and the post values from ```deployProxyURL``` before adding them to the properties file.
If this is the case, I suggest simplifying the implementation of this function and doing it once, instead of twice. In other words, we can use the ```http.ProxyFromEnvironment``` API regardless of whether ```noProxy``` is defined or not.

Notice that the tests are also wrong, because when they set no_proxy, the no_proxy value is identical to the http_proxy value, and that's why no host and port are added to the properties file.
The tests should use no_proxy values with wildcards, and check that the no proxy filtering works properly.

Enquier: I agree that this could be simplified. The key element here is the introduction of the `http.ProxyFromEnvironment` API which introduces a robust “NoProxy” logic in that if it returns  nil, nil then you should not set the proxy flag and if it returns nil, <proxy_url> then you should set the proxy flag to that URL. 

As you suggest I could effectively pull most of this code out (I tried to keep the flow of the original code) and replace it with http.ProxyFromEnvironment. I think it also was this way due to the fact that I originally thought that it mattered if we set it for Deploy vs Resolve but after digging in I realized that Gradle handles that part itself.

Enquier: As for the tests there are two sets of tests. One set where the No Proxy is identical and one where they differ.

Enquier: We could add tests that test the full gambit of no_proxy rules, I figured that was best left to the `http.ProxyFromEnvironment` team/codebase instead I simply wanted a test that proved if the proxy was set or not depending on the variables we passed

artifactory/utils/buildinfoproperties_test.go
```diff
@@ -63,21 +70,55 @@ func TestCreateSimplePropertiesFileWithProxy(t *testing.T) {
...
+	setNoProxy(noproxy, t)
+	var propertiesFileConfig = map[string]string{
+		"artifactory.resolve.contextUrl": "http://some.url.com",
+		"artifactory.publish.contextUrl": "http://some.other.url.com",
+		"artifactory.deploy.build.name":  "buildName",
+	}
+	createSimplePropertiesFile(t, propertiesFileConfig)
+	setProxy(proxyOrg, t)
+	setNoProxy(noproxyOrg, t)
+}
```
Enquier, line 105: Here is a test where the proxy matches the noproxy

artifactory/utils/buildinfoproperties_test.go
```diff
@@ -63,21 +70,55 @@ func TestCreateSimplePropertiesFileWithProxy(t *testing.T) {
...
+
```
Enquier, line 121: This test has the proxy and the no proxy not matching and should return proxy/port

eyalbe4:
Thanks @Enquier. Let me know after you add the changes and would like me to review this PR again.

eyalbe4:
Hi @Enquier,
Are you planning to continue working on this?

Enquier:
@eyalbe4 ive been tied up deploying Eplus internally I’ll take a look to see if I can resurrect this

PR #733, opened 2020-06-09
Create backup in config conversion

RobiNino:


utils/config/config.go
```diff
@@ -329,6 +332,11 @@ func convertIfNeeded(content []byte) ([]byte, error) {
 		}
 		fallthrough
 	case "1":
+		log.Debug("Converting JFrog config to the latest version...")
```
eyalbe4, line 335: I suggest changig this to:
log.Info("Converting JFrog CLI's config to the latest version...")

Issue #716, opened 2020-05-28
JFrog rt pip-install does not record downloaded dependencies


0xDAFE:
**Description**
jfrog-cli does not collect Python dependencies although downloaded from Artifactory server.

It complains:
> The pypi packages above could not be found in Artifactory or were not downloaded in this execution

However, the output shows that the artifacts are indeed downloaded from our Artifactory server.

**To Reproduce**
Basically follow the example here: [https://github.com/jfrog/project-examples/tree/master/python-example](https://github.com/jfrog/project-examples/tree/master/python-example)

- Configure jfrog-cli with your Artifactory instance and repos
- Create a fresh venv
- Activate the venv
- Install dependencies 

**Expected behavior**
jfrog-cli should record downloaded dependencies for further usage (e.g. XRay)

**Log**
```
jfrog rt pipi -r requirements.txt --build-name=vulnapp-build --build-number=1 --module=vulnapp --no-cache-dir --force-reinstall
[Info] Running pip Install.
Looking in indexes: https://myuser:****@artifactory.mydomain.com/artifactory/api/pypi/pypi-remote/simple
Collecting asgiref==3.2.7 (from -r requirements.txt (line 1))
  Downloading https://artifactory.mydomain.com/artifactory/api/pypi/pypi-remote/packages/packages/68/00/25013f7310a56d17e1ab6fd885d5c1f216b7123b550d295c93f8e29d372a/asgiref-3.2.7-py2.py3-none-any.whl
Collecting attrs==19.3.0 (from -r requirements.txt (line 2))
  Downloading https://artifactory.mydomain.com/artifactory/api/pypi/pypi-remote/packages/packages/a2/db/4313ab3be961f7a763066401fb77f7748373b6094076ae2bda2806988af6/attrs-19.3.0-py2.py3-none-any.whl
Collecting Django==3.0.6 (from -r requirements.txt (line 3))
  Downloading https://artifactory.mydomain.com/artifactory/api/pypi/pypi-remote/packages/packages/9d/04/04abb097c84c770180eeebe7ed920ce42f9917ab5ad4de01ff8ed11bc25b/Django-3.0.6-py3-none-any.whl (7.5MB)
Collecting feedgen==0.5.1 (from -r requirements.txt (line 4))
[...]

PR #747, opened 2020-06-20
Pip command - Search dependencies in remote cache in case of resolution from a remote repository.

liron-shalom:


liron-shalom:
fix issue #716

PR #749, opened 2020-06-23
Make refreshable tokens true by default

RobiNino:
Depends on [176](https://github.com/jfrog/jfrog-client-go/pull/176)


artifactory/cli.go
```diff
@@ -3642,8 +3649,7 @@ func offerConfig(c *cli.Context) (*config.ArtifactoryDetails, error) {
 	if err != nil {
 		return nil, err
 	}
-	encPassword := c.BoolT("enc-password")
-	configCmd := commands.NewConfigCommand().SetDefaultDetails(details).SetInteractive(true).SetEncPassword(encPassword)
+	configCmd := commands.NewConfigCommand().SetDefaultDetails(details).SetInteractive(true).SetEncPassword(true)
```
eyalbe4, line 3652: What is the reason that ```enc-password``` is now always true? Doesn't this mean that you won't be able to configure an Artifactory instance, which is not accessible at the time of the config?

utils/config/tokenrefresh.go
```diff
@@ -55,12 +55,12 @@ func tokenRefreshHandler(currentAccessToken string) (newAccessToken string, err
 	}
 
 	serverConfiguration, err := GetArtifactorySpecificConfig(tokenRefreshServerId, true, false)
```
eyalbe4, line 57: Let's add a description above the definition of the ```tokenRefreshServerId``` variable (at the top of the file).

artifactory/cli.go
RobiNino: Since the offer config is used in generic commands, the enc-password option is not exposed. I think exposing it makes no sense. The functionality so far was reading the flag, but since it is not exposed the default value was always returned.

utils/config/tokenrefresh.go
RobiNino: Done

PR #759, opened 2020-06-30
Exclude distributions from refreshable tokens

RobiNino:


artifactory/commands/config.go
```diff
@@ -465,8 +465,8 @@ func ClearConfig(interactive bool) {
 	config.SaveArtifactoryConf(make([]*config.ArtifactoryDetails, 0))
 }
 
-func GetConfig(serverId string) (*config.ArtifactoryDetails, error) {
-	return config.GetArtifactorySpecificConfig(serverId, true, false)
+func GetConfig(serverId string, distribution bool) (*config.ArtifactoryDetails, error) {
```
eyalbe4, line 468: I suggest renaming "distribution" to "excludeRefreshableTokens".

artifactory/cli.go
```diff
@@ -1890,7 +1890,7 @@ func getPipInstallFlags() []cli.Flag {
 }
 
 func createArtifactoryDetailsByFlags(c *cli.Context, distribution bool) (*config.ArtifactoryDetails, error) {
```
eyalbe4, line 1892: I suggest renaming "distribution" to "excludeRefreshableTokens".

PR #760, opened 2020-06-30
Exclude distribution from refresh tokens, improve Artifactory details collection

RobiNino:


artifactory/cli.go
```diff
@@ -3677,16 +3674,42 @@ func offerConfig(c *cli.Context) (*config.ArtifactoryDetails, error) {
...
+	createdDetails, err := offerConfig(c)
+	if err != nil {
+		return nil, err
+	}
+	if createdDetails != nil {
+		return createdDetails, err
+	}
+
+	details := createArtifactoryDetailsFromOptions(c)
+	// If urls or credentials were passed in options, use options as they are.
```
eyalbe4, line 3687: ...passed in options... --> ...passed as options...

Let's also add the following to the end of the description:

For security reasons, we'd like to avoid using part of the connection details from command options and the rest from the config. Either use command options only or config only.

PR #775, opened 2020-07-13
CI tests with JFrog Pipelines

yahavi:
-----

### DIFF
--- a/pipelines.yml
+++ b/pipelines.yml
@@ -1,0 +1,173 @@
+# Set common JFrog integration and GitHub input source
+commonConfiguration: &COMMON_CONFIGURATION
+  integrations:
+    - name: jfrog
+  inputResources:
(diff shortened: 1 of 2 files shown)


pipelines.yml
```diff
@@ -0,0 +1,173 @@
...
+      runtime:
+        type: image
+        image:
+          custom:
+            name: releases-docker.jfrog.io/jfrog-ecosystem-integration-env
+            tag: 1.0.0
+      environmentVariables:
+        readOnly:
+          GOPROXY: https://gocenter.io
+          JFROG_CLI_OFFER_CONFIG: "false"
```
eyalbe4, line 40: I think that this env var is no longer required, because when the CI env var is set true, all interactive prompts are disabled anyway.
I think we should add a task for deprecating JFROG_CLI_OFFER_CONFIG.

PR #792, opened 2020-07-27
New 'go get' install script

eyalbe4:
-----


build/goGet.sh
```diff
@@ -0,0 +1,9 @@
+GO111MODULE=on go get github.com/jfrog/jfrog-cli;
```
barbelity, line 1: I suggest adding '#!/bin/bash' at the beginning.

eyalbe4: @barbelity,
Since the code in this script in meant to be copied and then pasted in the terminal, through the copy buttons in https://jfrog.com/getcli/ I actually think is is best not to add the '#!/bin/bash' at the beginning.
What do you think?

barbelity: In this case you are right 👍

eyalbe4: What do you mean @barbelity?
What should we add to the script?

PR #823, opened 2020-09-08
Set/Delete props unified errors output

talarian1:
Set props command could have returned a none (nill) error instead of the totalLengthErr
that occurred.
The failCount of delete props command was not set in case of returning an error.
For consistency it is now set before the return error statement like in the set props command logic.

-----

### DIFF
--- a/artifactory/commands/generic/setprops.go
+++ b/artifactory/commands/generic/setprops.go
@@ -32,6 +32,9 @@ func (setProps *SetPropsCommand) Run() error {
 	}
 
 	reader, searchErr := searchItems(setProps.Spec(), servicesManager)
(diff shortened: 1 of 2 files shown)


artifactory/commands/generic/setprops.go
```diff
@@ -32,6 +32,9 @@ func (setProps *SetPropsCommand) Run() error {
 	}
 
 	reader, searchErr := searchItems(setProps.Spec(), servicesManager)
+	if searchErr != nil {
```
Or-Gabay, line 35: "searchErr" -> "err" (let's change in the deleteprops.go as well).

PR #831, opened 2020-09-14
Add Plugins Support

RobiNino:
Depends on [PR](https://github.com/jfrog/jfrog-cli-core/pull/3)


plugins/signatureutils.go
```diff
@@ -0,0 +1,109 @@
...
+func (signatureCmd *SignatureCmd) GetStdWriter() io.WriteCloser {
+	return nil
+}
+
+func (signatureCmd *SignatureCmd) GetErrWriter() io.WriteCloser {
+	return nil
+}
+
+// Gets all the installed plugins' signatures by looping over the plugins dir.
+func getPluginsSignatures() ([]*components.PluginSignature, error) {
```
eyalbe4, line 45: I think we should a test for this function.

eyalbe4 (changes requested):

PR #844, opened 2020-09-23
New build append command

yahavi:
-----

* Add build append command:
```
jfrog rt build-append <build name> <build number> <build name to append> <build number to append>
jfrog rt ba <build name> <build number> <build name to append> <build number to append>
```

* Add module type to npm and generic commands

Depends on: 
https://github.com/jfrog/jfrog-client-go/pull/205
https://github.com/jfrog/jfrog-cli-core/pull/12


buildinfo_test.go
```diff
@@ -230,6 +230,53 @@ func getFilesFromBuildDir(t *testing.T, buildName, buildNumber string) []os.File
...
+	buildInfo, _, err := tests.GetBuildInfo(artifactoryDetails, tests.RtBuildName2, buildNumber2)
+	assert.NoError(t, err)
+	assert.NotNil(t, buildInfo)
+	assert.Len(t, buildInfo.BuildInfo.Modules, 1)
+	module := buildInfo.BuildInfo.Modules[0]
+	assert.Equal(t, tests.RtBuildName1+"/"+buildNumber1, module.Id)
+
+	// Clean builds
+	inttestutils.DeleteBuild(artifactoryDetails.Url, tests.RtBuildName1, artHttpDetails)
+	inttestutils.DeleteBuild(artifactoryDetails.Url, tests.RtBuildName2, artHttpDetails)
```
eyalbe4, line 277: I think we should call ```cleanArtifactoryTest()``` at the end fo this test. I actually noticed that there are a few other tests in this file which do not include this. Maybe we should add it to the end of all tests.

eyalbe4 (changes requested):

inttestutils/buildinfo.go
```diff
@@ -28,3 +35,18 @@ func DeleteBuild(artifactoryUrl, buildName string, artHttpDetails httputils.Http
 		log.Error(string(body))
 	}
 }
+
+func ValidateGeneratedBuildInfo(t *testing.T, buildname, buildNumber string, moduleNames []string, moduleType buildinfo.ModuleType) {
```
eyalbe4, line 39: U think that this function may fail, if it used on a build-info which includes multiple modules, with different modules types. This is because of this assertion:
```
assert.Equal(t, moduleType, module.Type)
```
I  think we should replace this assertion with an addition type check as part of the if.
Also, we may want to change the name of the function from ```ValidateGeneratedBuildInfo``` to ```ValidateGeneratedBuildInfoModule``` to better describe what the function does.

Since the validation can fail due to the type as well, let's add the module type to the the following message:
```
assert.Fail(t, fmt.Sprintf("Module %s is expected to be one of %v", module.Id, moduleNames))
```

buildinfo_test.go
```diff
@@ -230,6 +230,54 @@ func getFilesFromBuildDir(t *testing.T, buildName, buildNumber string) []os.File
...
+	cleanArtifactoryTest()
```
eyalbe4, line 278: Let's add ```cleanArtifactoryTest()``` to the other functions which don't have it. For example, ```TestBuildAddDependencies```

docker_test.go
```diff
@@ -54,9 +55,11 @@ func runDockerPushTest(imageName, module string, withModule bool, t *testing.T)
 	// Push docker image using docker client
 	if withModule {
 		artifactoryCli.Exec("docker-push", imageTag, *tests.DockerTargetRepo, "--build-name="+tests.DockerBuildName, "--build-number="+buildNumber, "--module="+module)
+		inttestutils.ValidateGeneratedBuildInfo(t, tests.DockerBuildName, buildNumber, []string{module}, buildinfo.Docker)
```
eyalbe4, line 58: Let's add the above validation also to the ```else``` section of this ```if```. We'll need to use the default module name value there.

eyalbe4 (changes requested):

PR #845, opened 2020-09-29
Support plugin installation

RobiNino:
Depends on [PR](https://github.com/jfrog/jfrog-client-go/pull/206) & [PR](https://github.com/jfrog/jfrog-cli-core/pull/15)

### DIFF
--- a/docs/plugin/install/help.go
+++ b/docs/plugin/install/help.go
@@ -1,0 +1,8 @@
+package install
+
+const Description = "Install plugin."
+
+var Usage = []string{"jfrog plugin install <plugin name>"}
+
+const Arguments string = `	plugin name
+		Specifies the name of the JFrog CLI Plugin you wish to install from the plugins registry.`
--- a/docs/plugin/install/help.go
+++ b/docs/plugin/install/help.go
@@ -1,8 +1,10 @@
 package install
 
-const Description = "Install plugin."
+const Description = "Install a JFrog CLI plugin."
 
-var Usage = []string{"jfrog plugin install <plugin name>"}
+var Usage = []string{"jfrog plugin install <plugin name and version>"}
 
-const Arguments string = `	plugin name
-		Specifies the name of the JFrog CLI Plugin you wish to install from the plugins registry.`
+const Arguments string = `	plugin name and version
+		Specifies the name and version of the JFrog CLI Plugin you wish to install from the plugins registry.
+		The version should be specified after a '@' separator, such as: 'hello-frog@1.0.0'. 
+		To download the latest version, specify the plugin name only.`
--- a/docs/plugin/uninstall/help.go
+++ b/docs/plugin/uninstall/help.go
@@ -1,6 +1,6 @@
 package install
 
-const Description = "Uninstall plugin."
+const Description = "Uninstall a JFrog CLI plugin."
 
 var Usage = []string{"jfrog plugin uninstall <plugin name>"}
 
--- a/docs/plugin/uninstall/help.go
+++ b/docs/plugin/uninstall/help.go
@@ -1,0 +1,8 @@
+package install
+
+const Description = "Uninstall plugin."
+
+var Usage = []string{"jfrog plugin uninstall <plugin name>"}
+
+const Arguments string = `	plugin name
+		Specifies the name of the JFrog CLI Plugin you wish to uninstall from the local plugins pool.`
--- a/plugins/commands/uninstall.go
+++ b/plugins/commands/uninstall.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"github.com/codegangsta/cli"
 	"github.com/jfrog/jfrog-cli-core/utils/coreutils"
+	pluginsutils "github.com/jfrog/jfrog-cli/plugins/utils"
 	"github.com/jfrog/jfrog-cli/utils/cliutils"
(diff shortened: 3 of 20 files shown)


docs/plugin/install/help.go
```diff
@@ -0,0 +1,8 @@
+package install
+
+const Description = "Install plugin."
```
eyalbe4, line 3: Install plugin. --> Install a JFrog CLI plugin.

docs/plugin/uninstall/help.go
```diff
@@ -0,0 +1,8 @@
+package install
+
+const Description = "Uninstall plugin."
```
eyalbe4, line 3: Uninstall plugin. --> Uninstall a JFrog CLI plugin.

docs/plugin/install/help.go
```diff
@@ -0,0 +1,8 @@
...
+
+var Usage = []string{"jfrog plugin install <plugin name>"}
+
+const Arguments string = `	plugin name
```
eyalbe4, line 7: Users should be abel to install a specific plugin version.
For example, to install version 1.0.0 of my-plugin:
```
jfrog plugin install my-plugin@1.0.0
```

eyalbe4 (changes requested):
I haven't finished reviewing everything, but I'm releasing what I have so far.

plugins/commands/uninstall.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	for _, plugin := range plugins {
+		if strings.EqualFold(plugin, requestedPlugin) {
+			ci, err := clientutils.GetBoolEnvValue(coreutils.CI, false)
+			if err != nil {
+				return err
+			}
+			if !ci {
+				if !coreutils.AskYesNo("Are you sure you want to delete plugin: \""+plugin+"\"?", false) {
```
eyalbe4, line 44: delete --> uninstall

plugins/commands/uninstall.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func UninstallCmd(c *cli.Context) error {
+	if c.NArg() != 1 {
+		return cliutils.PrintHelpAndReturnError("Wrong number of arguments.", c)
+	}
+	return runUninstallCmd(c.Args().Get(0))
+}
+
+func runUninstallCmd(requestedPlugin string) error {
```
eyalbe4, line 22: What is the reason for scanning the plugins directory twice?
We're scanning it once with the ```utils.GetAllPluginsNames()``` function and then again with the for loop in this function.
Can't we just find the local path of the plugin binary using the provided plugin name?

plugins/utils/signatureutils.go
```diff
@@ -44,45 +45,49 @@ func (signatureCmd *SignatureCmd) GetErrWriter() io.WriteCloser {
...
 	var signatures []*components.PluginSignature
-	pluginsDir, err := coreutils.GetJfrogPluginsDir()
+	pluginsNames, err := GetAllPluginsNames()
 	if err != nil {
-		return signatures, err
-	}
-	exists, err := fileutils.IsDirExists(pluginsDir, false)
-	if err != nil || !exists {
-		return signatures, err
+		return signatures, errorutils.CheckError(err)
```
eyalbe4, line 50: No need for ```CheckError``` here, because it the error is already checked inside ```GetAllPluginsNames()```.

plugins/utils/signatureutils.go
```diff
@@ -92,10 +97,14 @@ func signaturesToCommands(signatures []*components.PluginSignature) []cli.Comman
 			Usage:           sig.Usage,
 			SkipFlagParsing: true,
 			Action: func(c *cli.Context) error {
-				return gofrogcmd.RunCmd(
+				output, err := gofrogcmd.RunCmdOutput(
 					&SignatureCmd{
 						sig.ExecutablePath,
 						c.Args()})
+				if err == nil {
+					log.Output(output)
```
eyalbe4, line 105: We probably don't want this output as debug only.

utils/tests/utils.go
```diff
@@ -567,3 +568,35 @@ func CleanUpOldItems(baseItemNames []string, getActualItems func() ([]string, er
 		}
 	}
 }
+
+// Set HomeDir to desired location.
+// Caller is responsible to set the old home location back.
+func SetJfrogHome(newHome string) (oldHome string, err error) {
```
eyalbe4, line 574: I noticed that in ```TestUnitTests``` this function is used, without a defer clause that reverts back to the old home dir. Let's add it.

eyalbe4 (changes requested):
Done reviewing everything.

docs/plugin/install/help.go
RobiNino: Added an explanation. Please tell me what you think

plugins/commands/uninstall.go
RobiNino: I thought to support the command case-insensitive but I agree with you, and I think we should make all plugins names lowercase as a constraint.

plugins/utils/signatureutils.go
RobiNino: I am not sure I understand what you meant here. The purpose of this log is to set the output of the plugin command, as output of the CLI process, so users will be able to catch it's output.

eyalbe4 (approved):

PR #852, opened 2020-10-06
Bug fix - Displayed summary is cut

eyalbe4:
-----
This PR depends on https://github.com/jfrog/jfrog-cli-core/pull/16 to be first merged.
The details of the issue are documented here - https://github.com/jfrog/jfrog-cli-core/pull/16


RobiNino (changes requested):
I suggest leaving the `progressBar.quit()` in the cli. Moving it to the `Run` function might cause problems, such as races between the logs of the report usage to the closing of the progress bar.
It is also good practice that the initialiser is also responsible to closing.

I think we can wrap the Exec function here, and set the defers inside that wrap function.
Note the `defer logUtils.CloseLogFile(downloadCommand.LogFile())` should be called in the same place and manner as the `Quit`, as it is also responsible for setting the older logger back.

eyalbe4:
Done @RobiNino.
This PR depends on https://github.com/jfrog/jfrog-cli-core/pull/18.

RobiNino (approved):

PR #865, opened 2020-10-16
Make username optional in access-token-create

yahavi:
-----

`jfrog rt atc` - If username is not specified, use the current user

### DIFF
--- a/artifactory/cli.go
+++ b/artifactory/cli.go
@@ -2708,7 +2708,7 @@ func accessTokenCreateCmd(c *cli.Context) error {
 		return err
 	}
 
-	if c.NArg() != 1 {
+	if c.NArg() > 1 {
 		return cliutils.PrintHelpAndReturnError("Wrong number of arguments.", c)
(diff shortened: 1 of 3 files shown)


artifactory/cli.go
```diff
@@ -2708,20 +2708,26 @@ func accessTokenCreateCmd(c *cli.Context) error {
...
+	if c.NArg() > 1 {
 		return cliutils.PrintHelpAndReturnError("Wrong number of arguments.", c)
 	}
 
 	rtDetails, err := createArtifactoryDetailsByFlags(c, false)
 	if err != nil {
 		return err
 	}
+	var userName string
+	if c.NArg() > 0 {
```
eyalbe4, line 2720: Let's add the following comment above this line:
// If the username is provided as an argument, then it is used when creating the token. If not, then the configured username (or the the value of the --user option) is used.

eyalbe4 (changes requested):

PR #875, opened 2020-10-27
fix: handle insecure-tls in all commands

fsamin:
I wasn't able to run the `access-token-create` command with the `--insecure-tls` flag.
So I suggest to handle the `insecure-tls` in all commands. I confirm that the lower layers of the CLI handle it perfectly.

Signed-off-by: francois  samin <francois.samin@corp.ovh.com>

-----

### DIFF
--- a/utils/cliutils/commandsflags.go
+++ b/utils/cliutils/commandsflags.go
@@ -1,10 +1,11 @@
 package cliutils
 
 import (
-	"github.com/codegangsta/cli"
-	"github.com/jfrog/jfrog-client-go/utils/log"
 	"sort"
 	"strconv"
+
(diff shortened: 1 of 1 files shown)


fsamin:

fsamin:
recheckcla

eyalbe4:
@fsamin,
Thanks for creating this PR.
Just adding the ```--insecure-tls``` flag to those commands will not make it work. This is because the implementation does not use it. Unlike the ```mvn``` command, which translates this flag into maven equivalent maven options, ```gradle``` or the ```docker``` commands do not.

eyalbe4:
@fsamin,
Please let me know if you'd like to keep this PR open.

PR #876, opened 2020-10-28
Add cleanGoCache to the beginning of every go test

RobiNino:


eyalbe4 (changes requested):
Here's an idea that can help developers not to forget the clean up before running go.
Why don't wrap ```artifactoryCli.Exec``` with a new ```execGo``` function inside ```go_test.go```. This function will take care of running ```cleanGoCache(t)``` before running the go command.

PR #893, opened 2020-11-10
Use GetBuildInfo in tests

RobiNino:
Depends on [jfrog-cli-core/pull/36](https://github.com/jfrog/jfrog-cli-core/pull/36)


utils/tests/utils.go
```diff
@@ -302,6 +304,44 @@ func DeleteFiles(deleteSpec *spec.SpecFiles, artifactoryDetails *config.Artifact
...
+		return buildinfo.BuildInfo{}, err
+	}
+	if notFound {
+		assert.False(t, notFound, "expected to find build info")
+		return buildinfo.BuildInfo{}, errors.New("expected to find build info")
+	}
+	return bi.BuildInfo, nil
+}
+
+func VerifyBuildInfoDoesntExist(t *testing.T, artDetails *config.ArtifactoryDetails, buildName, buildNumber string) (doesntExist bool, err error) {
```
eyalbe4, line 328: This function verifies that the build info requested does not exist. Looking at this function's two usages, they also do the same check, by using the *doesntExist* returned value. I actually suggest merging this function with its neighbouring *GetBuildInfo* function. The *GetBuildInfo* can simply return the *doesntExist* boolean, and not do any assets. The calling functions can do the assets.

buildinfo_test.go
```diff
@@ -238,19 +248,31 @@ func TestBuildAddDependencies(t *testing.T) {
...
+func testArtifactoryPublishWithoutBuildUrl(t *testing.T, buildName, buildNumber string) {
 	initArtifactoryTest(t)
-	buildNumber := "10"
-	inttestutils.DeleteBuild(artifactoryDetails.Url, tests.RtBuildName1, artHttpDetails)
+	inttestutils.DeleteBuild(artifactoryDetails.Url, buildName, artHttpDetails)
 
-	body := uploadFilesAndGetBuildInfo(t, tests.RtBuildName1, buildNumber, "")
+	bi, err := uploadFilesAndGetBuildInfo(t, buildName, buildNumber, "")
+	if err != nil {
+		return
```
eyalbe4, line 266: Would you like to add an assertNoError here, before the return?
The same goes for a few other places added by this PR.

go.mod
```diff
@@ -17,9 +17,9 @@ require (
 	gopkg.in/yaml.v2 v2.2.2
 )
 
-// replace github.com/jfrog/jfrog-client-go => github.com/jfrog/jfrog-client-go dev
+replace github.com/jfrog/jfrog-client-go => /Users/robin/proj-dev/jfrog-client-go
 
-// replace github.com/jfrog/jfrog-cli-core => github.com/jfrog/jfrog-cli-core dev
+replace github.com/jfrog/jfrog-cli-core => /Users/robin/proj-dev/jfrog-cli-core
```
eyalbe4, line 22: Let's remove before merging.

eyalbe4 (changes requested):

buildinfo_test.go
RobiNino: The function asserts the before returning the error. I didn't assert again to avoid duplication. Do you think it is necessary?

RobiNino: Since this function is used by only a few similar tests it asserts inside. I added assertions everywhere else

PR #917, opened 2020-12-06
Allow args with spaces in NuGet / Dotnet restore

RobiNino:
Depends on [jfrog/jfrog-cli-core#42](https://github.com/jfrog/jfrog-cli-core/pull/42)


artifactory/cli.go
```diff
@@ -1317,7 +1317,14 @@ func nugetLegacyCmd(c *cli.Context) error {
...
 		return err
 	}
-	nugetCmd.SetBasicCommand(c.Args().Get(0)).SetArgAndFlags(c.String("nuget-args")).
+
+	args := c.String("nuget-args")
+	if args == "" {
+		nugetCmd.SetArgAndFlags([]string{})
+	} else {
+		nugetCmd.SetArgAndFlags(strings.Split(args, " "))
+	}
```
eyalbe4, line 1326: Let's move the above if / else statement into a ```toArgsList``` function, and make this function a bit shorter.

eyalbe4 (changes requested):

PR #920, opened 2020-12-08
Add NuGet v3 support

### LINKED ISSUE #881: JFrog CLI doesn't support V3 Nugget packages while Artifactory does.

RobiNino:
Fixes #881.
Depends on [jfrog/jfrog-cli-core#45](https://github.com/jfrog/jfrog-cli-core/pull/45).

NuGet v3 will now be the default protocol. 
Using v2 can be configured in the project config file (by running `nugetc` or `dotnetc`, with `--nuget-v2` or interactively), or by option for the legacy nuget command (`--nuget-v2-protocol`).

### DIFF
--- a/artifactory/cli.go
+++ b/artifactory/cli.go
@@ -1314,11 +1314,11 @@ func nugetLegacyCmd(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	nugetCmd.SetBasicCommand(c.Args().Get(0)).SetArgAndFlags(c.String("nuget-args")).
+	nugetCmd.SetBasicCommand(c.Args().Get(0)).SetArgAndFlags(c.String(cliutils.NugetArgs)).
 		SetRepoName(c.Args().Get(1)).
 		SetBuildConfiguration(buildConfiguration).
-		SetSolutionPath(c.String("solution-root")).
-		SetUseNugetV2(c.Bool("nuget-v2-protocol")).
+		SetSolutionPath(c.String(cliutils.SolutionRoot)).
+		SetUseNugetV2(c.Bool(cliutils.LegacyNugetV2)).
 		SetRtDetails(rtDetails)
 
 	return commands.Exec(nugetCmd)
--- a/artifactory/cli.go
+++ b/artifactory/cli.go
@@ -1267,12 +1267,8 @@ func nugetCmd(c *cli.Context) error {
 		if c.NArg() < 1 {
 			return cliutils.PrintHelpAndReturnError("Wrong number of arguments.", c)
 		}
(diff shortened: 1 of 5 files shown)


artifactory/cli.go
```diff
@@ -1321,6 +1318,7 @@ func nugetLegacyCmd(c *cli.Context) error {
 		SetRepoName(c.Args().Get(1)).
 		SetBuildConfiguration(buildConfiguration).
 		SetSolutionPath(c.String("solution-root")).
+		SetUseNugetV2(c.Bool("nuget-v2-protocol")).
```
eyalbe4, line 1321: nuget-v2-protocol --> nuget-v2
Since we have the ```legacyNugetV2``` constant, let's change it there, and use it instead of the string. Actually, this is something we should do for all flags (all commands).
Actually - do we need both ```legacyNugetV2``` and ```nugetV2```?

artifactory/cli.go
```diff
@@ -1374,6 +1375,24 @@ func dotnetCmd(c *cli.Context) error {
 	return commands.Exec(dotnetCmd)
 }
 
+func getNugetAndDotnetConfigFields(configFilePath string) (rtDetails *config.ArtifactoryDetails, targetRepo string, useNugetV2 bool, err error) {
```
eyalbe4, line 1378: I noticed that only for the nuget and dotnet commands, the config is read here, and not in jfrog-cli-core. All other commands use the ```SetConfigFilePath``` API when creating the command.
What is the reason this has been now modified for nuget and dotnet? I'm not sure which method is better. My concern is more around consistency between all c ommands.

go.mod
```diff
@@ -19,7 +19,7 @@ require (
 
 // replace github.com/jfrog/jfrog-client-go => github.com/jfrog/jfrog-client-go v0.16.0
 
-// replace github.com/jfrog/jfrog-cli-core => github.com/jfrog/jfrog-cli-core v1.1.2-0.20201118101632-f498d864d81b
+replace github.com/jfrog/jfrog-cli-core => github.com/RobiNino/jfrog-cli-core nuget-v3
```
eyalbe4, line 22: Let's change this before committing.

eyalbe4 (changes requested):

artifactory/cli.go
RobiNino: Both are needed because the flags have different settings (the legacy is hidden). The flags map cannot contain both with the same key, hence the different name. I cannot rename it to nuget-v2, do you have a different naming suggestion?

artifactory/cli.go
RobiNino: The legacy nuget command doesn't use a config file so it doesn't really make sense to read it in the core command.

PR #952, opened 2021-01-06
Add docker tests for virtual repositories

Or-Geva:
-----

### DIFF
--- a/docker_test.go
+++ b/docker_test.go
@@ -45,44 +45,49 @@ func initContainerTest(t *testing.T) []container.ContainerManagerType {
 
 func TestContainerPush(t *testing.T) {
 	containerManagers := initContainerTest(t)
-	for _, containerManager := range containerManagers {
-		runPushTest(containerManager, tests.DockerImageName, tests.DockerImageName+":1", false, t)
+	for _, repo := range []string{*tests.DockerVirtualRepo, *tests.DockerTargetRepo} {
+		for _, containerManager := range containerManagers {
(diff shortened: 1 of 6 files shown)


docker_test.go
```diff
@@ -96,46 +101,48 @@ func TestContainerPushBuildNameNumberFromEnv(t *testing.T) {
...
+			validateContainerBuild(tests.DockerBuildName, buildNumber, imagePath, ModuleNameJFrogTest, 0, 7, 7, t)
 
-		inttestutils.ContainerTestCleanup(t, artifactoryDetails, artHttpDetails, tests.DockerImageName, tests.DockerBuildName)
+			inttestutils.ContainerTestCleanup(t, artifactoryDetails, artHttpDetails, tests.DockerImageName, tests.DockerBuildName, repo)
+		}
 	}
 }
 
-func containerTestCleanup(t *testing.T, imageName, buildName string) {
+func containerTestCleanup(t *testing.T, imageName, buildName, repo string) {
```
eyalbe4, line 142: There are a f ew functions in this file, which publish build-info, but instead of invoking this function, it invokes ```inttestutils.ContainerTestCleanup``` directly, and the therefore, the published build doesn't get deleted.

eyalbe4 (changes requested):

docker_test.go
Or-Geva: No calls are made to this function (probably old).
This function `inttestutils.ContainerTestCleanup` is getting called by all docker tests and it includes the deletion of the published build.

PR #960, opened 2021-01-25
Support path mapping in distribution

yahavi:
-----

Depends on: https://github.com/jfrog/jfrog-client-go/pull/274

Add `--target` flag to the `release-bundle-create` and `release-bundle-update` commands to support Distribution path mapping. For more information about the file-spec-based path mapping see https://github.com/jfrog/jfrog-client-go/pull/274.

### DIFF
--- a/utils/cliutils/commandsflags.go
+++ b/utils/cliutils/commandsflags.go
@@ -322,6 +322,7 @@ const (
 	rbDryRun            = releaseBundlePrefix + dryRun
 	rbRepo              = releaseBundlePrefix + repo
 	rbPassphrase        = releaseBundlePrefix + passphrase
+	distTarget          = releaseBundlePrefix + target
 	sign                = "sign"
 	desc                = "desc"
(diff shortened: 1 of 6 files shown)


utils/cliutils/commandsflags.go
```diff
@@ -997,6 +998,11 @@ var flagsMap = map[string]cli.Flag{
 		Name:  passphrase,
 		Usage: "[Optional] The passphrase for the signing key. ` `",
 	},
+	distTarget: cli.StringFlag{
+		Name: target,
+		Usage: "[Optional] The target path for artifacts matched in the pattern path. " +
```
eyalbe4, line 1003: The target path for artifacts matched in the pattern path. --> The target path for distributed artifacts on the edge node. If not specified, the artifacts will have the same path and name on the edge node, as on the source Artifactory server.

eyalbe4 (changes requested):

PR #970, opened 2021-02-08
Support for Artifactory Projects

eyalbe4:
-----
This PR adds support for the new Projects functionality in Artifactory. It relies on the following PRs to be merged before this one:
https://github.com/jfrog/jfrog-cli-core/pull/66
https://github.com/jfrog/jfrog-client-go/pull/283

1. The build-publish command now accpets an optional "project" flag, with the project key value. If provided, the build-info is published to the project's build-info repository, named my-project-key-build-info. If not provided, the build-info is published (by default) to a repository named artifactory-build-info.

2. All commands which deploy files to Artifactory also accpet the new "project" flag. They use it to add a new "build.repo" property on each of the deployed artifacts, in addition to the existing "build.name", "build.number" and "build.timestamp" props. The "build.repo" prop stores the name of the build-info repository, which is part of the project, to which the artifacts and the build-info are deployed to.


yahavi (changes requested):
Let's add the `--project` to build-publish and build-discard.
I think we should also add the `--project` flag to all commands that create a build (like download, move, copy, ...) to allow saving the partial build info directory under `<buildName>_<buildNumber>_<project>`.

PR #992, opened 2021-03-08
Support private JFrog CLI plugins

RobiNino:


plugins/commands/uninstall_test.go
```diff
@@ -47,14 +47,17 @@ func TestRunUninstallCmd(t *testing.T) {
...
+	pluginExePath := filepath.Join(pluginsDir, utils.GetLocalPluginExecutableName(pluginName))
 	// Fix path for windows.
 	assert.NoError(t, os.Rename(filepath.Join(pluginsDir, pluginName), pluginExePath))
 
 	// Try uninstalling a plugin that doesn't exist.
 	err = runUninstallCmd("non-existing-plugin")
 	expectedError := generateNoPluginFoundError("non-existing-plugin")
 	assert.Error(t, err)
+	if expectedError == nil {
+		return
```
eyalbe4, line 59: Should this "return" statement be here? If so, worth adding a comment describing why.

eyalbe4 (changes requested):
Like the code!
Question related to the tests:
Does the user running the tests need to create the "jfrog-cli-plugins" repo manually before running the tests? If so, I think it is best that the test code creates such a repo automatically (with a name that is different thsn the default).
I think however this can be donw in a follow up PR. No need to delay this PR because of this.

plugins/commands/utils/utils.go
```diff
@@ -0,0 +1,130 @@
...
+	case "386":
+		return "linux-386", nil
+	case "s390x":
+		return "linux-s390x", nil
+	}
+	return "", errorutils.CheckError(errors.New("no compatible plugin architecture was found for the architecture of this machine"))
+}
+
+func CreatePluginsHttpDetails(rtDetails *config.ServerDetails) httputils.HttpClientDetails {
+	if rtDetails.AccessToken != "" && rtDetails.RefreshToken == "" {
```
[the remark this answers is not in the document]
RobiNino: Yes it does

plugins/commands/uninstall_test.go
RobiNino: It is not really needed because the function always returns an error, but the IDE warns about potential NPE.
Added a comment anyway

RobiNino:
The user does not need to create a repo for the tests, it is created by the tests suite.

PR #1037, opened 2021-04-01
Ci-setup: update minimal supported Xray version

talarian1:
-----


eyalbe4:
@talarian1,
After giving this more thought, I believe that the versions restrictions are actually not needed. The command can provide great value, even if used with old versions of Artifactory and Xray.
I opened https://github.com/jfrog/jfrog-cli/pull/1041 to remove the restrictions altogether. It would be great if you can review it.

PR #1046, opened 2021-04-05
Ci-Setup - Improvements

RobiNino:
1. If technologies built match history, offer the saved build command. Otherwise, generate a default one.
2. If no technology is chosen, return err.

Depends on [jfrog/jfrog-cli-core/pull/106](https://github.com/jfrog/jfrog-cli-core/pull/106/files)


main.go
```diff
@@ -141,18 +144,17 @@ func getCommands() []cli.Command {
 			Description: "Config commands",
 			Subcommands: config.GetCommands(),
 		},
-		// Disabled
```
eyalbe4, line 144: Let's re-comment-out this command.

eyalbe4 (changes requested):

PR #1049, opened 2021-04-08
integration tests improvements

yahavi:
-----

* Unset `JFROG_CLI_BUILD_NAME` and `JFROG_CLI_BUILD_NUMBER` environment variables before artifactory tests.
* Disallow chaining the timestamp to repo and build names more than once (`<repo-name>-<timestamp>-<timestamp>`).
* AQL Search results in TestArtifactoryUploadAsArchive are not necessarily ordered. Therefore we check them like that.
* Update go.mod and go.sum.
* Add cleanup after TestArtifactoryCreateUsers.


utils/tests/utils.go
```diff
@@ -476,6 +477,9 @@ func getSubstitutionMap() map[string]string {
 
 // Add timestamp to builds and repositories names
 func AddTimestampToGlobalVars() {
+	if timestampAdded {
```
talarian1, line 480: Do you think it is clear enough to someone that doesn't familiar with the original issue ? 
I added the following comment: 
`	// Make sure the global timestamp is added only once even in case of multiple  tests flags`

PR #1112, opened 2021-06-21
Move distribution to "jfrog ds"

gailazar300:
-----


distribution/cli.go
```diff
@@ -0,0 +1,334 @@
...
+	if !(c.NArg() == 2 && c.IsSet("spec") || (c.NArg() == 3 && !c.IsSet("spec"))) {
+		return cliutils.PrintHelpAndReturnError("Wrong number of arguments.", c)
+	}
+	if c.IsSet("detailed-summary") && !c.IsSet("sign") {
+		return cliutils.PrintHelpAndReturnError("The --detailed-summary option can't be used without --sign", c)
+	}
+	var releaseBundleCreateSpec *spec.SpecFiles
+	var err error
+	if c.IsSet("spec") {
+		releaseBundleCreateSpec, err = artifactory.GetSpec(c, true)
```
eyalbe4, line 106: Now that the ```GetSpec``` is used by both ```cli.go``` of Artifactory and Distribution, the right thing to do in terms of design is to move it out of ```artifactory/cli.go``` into ```cliutils/utils.go```

artifactory/cli.go
```diff
@@ -1053,7 +979,7 @@ func GetCommands() []cli.Command {
 	})
 }
 
-func createArtifactoryDetailsByFlags(c *cli.Context, distribution bool) (*coreConfig.ServerDetails, error) {
+func CreateArtifactoryDetailsByFlags(c *cli.Context, distribution bool) (*coreConfig.ServerDetails, error) {
```
eyalbe4, line 982: Notice that this function includes ```if distribution```, although we're inside ```artifactory/cli.go```.
Here's what I suggest we do:
1. Make this function private again.
2. Create a similar function inside ```distribution/cli.go```, which includes only the distribution related code (without the if statement which is not needed anymore).
3. Also remove the if statement in the function in this file.

eyalbe4 (changes requested):

eyalbe4 (changes requested):
Looks like this PR was created on the master branch. Let's create it on the dev branch, so that it doesn't end up merged into the v1 release of jfrog-cli.

PR #1167, opened 2021-07-28
Refactor tests flags & add new Xray's test suite

talarian1:
-----
Change Artifactory authentication flags to JFrog platform authentication flags: rt.* -> jfrog.*
Add new Xray's integration tests.


pipelines.yml
```diff
@@ -120,8 +120,7 @@ pipelines:
           onExecute:
             - cd $res_jfrogCliGit_resourcePath
             - *CREATE_RT_ACCESS_TOKEN
-            - curl -X POST -u $int_jfrog_rt_user:$int_jfrog_rt_password $int_jfrog_dist_url/api/v1/security/token -d "username=$int_jfrog_rt_user" -d 'expires_in=1800' | jq '.access_token' -r > distToken
```
yahavi, line 123: This will probably not work. The Access token created is valid only for the Artifactory service.

pipelines.yml
```diff
@@ -169,4 +168,15 @@ pipelines:
...
+      - name: Xray
+        type: Bash
+        configuration:
+          <<: *COMMON_CONFIGURATION
+        execution:
+          <<: *UPDATE_COMMIT_STATUS
+          onExecute:
+            - cd $res_jfrogCliGit_resourcePath
+            - *CREATE_RT_ACCESS_TOKEN
+            - env -i PATH=$PATH M2_HOME=$M2_HOME HOME=$HOME `which go` test -v -timeout 0 -test.xray -jfrog.url=$int_jfrog_rt_url -jfrog.accessToken=$(< rtToken)
```
yahavi, line 182: This will probably not work. The Access token created is valid only for the Artifactory service.

yahavi (changes requested):

xray_test.go
```diff
@@ -0,0 +1,98 @@
...
+	if xrayCli != nil {
+		return
+	}
+	cred := authenticateXray()
+	xrayCli = tests.NewJfrogCli(execMain, "jfrog xr", cred)
+}
+
+// Tests basic binary scan by providing pattern (path to testdata binaries) and --licenses flag
+// and asserts any error.
+func TestXrayBinaryScan(t *testing.T) {
```
eyalbe4, line 67: I syggest that instead of merely tsting that the scan does not return an error, we run the scan with the ```--json``` option and see that we get a valid JSON response from Xray.

xray_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	xrayVersion, err := getXrayVersion()
+	if err != nil {
+		assert.NoError(t, err)
+		return
+	}
+	// If minimal version was supplied, make sure the Xray version fulfil the minimum version requirement
+	if len(minVersion) > 0 && !xrayVersion.AtLeast(minVersion[0]) {
+		t.Skip(fmt.Sprintf("Skipping Xray test. You are using Xray %s, while  this test requires Xray version %s or higher.", xrayVersion, minVersion))
+
```
eyalbe4, line 85: Redundant new-line here.

eyalbe4 (changes requested):
I think we should also add tests for the "audit-mvn", "audit-npm" and "audit-gradle" commands.
We can add it in a follow-up pull request if you prefer.

xray_test.go
```diff
@@ -0,0 +1,196 @@
...
+func TestXrayAuditNpm(t *testing.T) {
+	initXrayTest(t, xrutils.GraphScanMinVersion)
+	newStdout, stdWriter, previousStdout := tests.RedirectStdOutToPipe()
+	// Restore previous stdout when the function returns
+	defer func() {
+		os.Stdout = previousStdout
+	}()
+	tempDirPath, err := fileutils.CreateTempDir()
+	assert.NoError(t, err)
+	defer fileutils.RemoveTempDir(tempDirPath)
```
eyalbe4, line 98: The above line should be changed to:
```
defer assert.NoError(t, fileutils.RemoveTempDir(tempDirPath))
```
The code includes a few other places where ```RemoveTempDir``` doesn't include error check, so let's fix them as well.
The same goes for the ```defer os.Chdir(prevWd)``` line below in this function.

eyalbe4 (changes requested):

PR #1213, opened 2021-08-26
Fix Xray's tests output redirection

talarian1:
-----

### DIFF
--- a/xray_test.go
+++ b/xray_test.go
@@ -73,27 +73,14 @@ func initXrayCli() {
 // and asserts any error.
 func TestXrayBinaryScan(t *testing.T) {
 	initXrayTest(t, xrutils.GraphScanMinVersion)
-	newStdout, stdWriter, previousStdout := tests.RedirectStdOutToPipe()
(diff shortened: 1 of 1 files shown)


eyalbe4 (changes requested):
I'm wondering whether the ```RedirectLogOutputToBuffer``` utility which is already in use by some of there tests can't be used here. It's probably preferred to reuse utilities we already have, rather than creating new ones.

talarian1:
> I'm wondering whether the `RedirectLogOutputToBuffer` utility which is already in use by some of there tests can't be used here. It's probably preferred to reuse utilities we already have, rather than creating new ones.

The utility you mentioned is used to redirect the log output and not the standard output.
The new utility, added in #1167, is based on the `RedirectLogOutputToBuffer` but has a different purpose.
I don't like the idea to change the behavior of our logic, in order to test it.  But if the table/JSON is printed as a log message and not to the STD out we will be able to use only the `RedirectLogOutputToBuffer`.

PR #1231, opened 2021-09-22
Use project flag in search/upload/delete/properties commands

sverdlov93:
-----


artifactory/cli.go
```diff
@@ -2846,6 +2850,7 @@ func createDefaultUploadSpec(c *cli.Context) (*spec.SpecFiles, error) {
 		Props(c.String("props")).
 		TargetProps(c.String("target-props")).
 		Build(c.String("build")).
+		Project(c.String("project")).
```
eyalbe4, line 2853: I think that both the "project" and "build" file spec properties shouldn't be included as part of the upload spec.
This is because both are used for filtering out files on Artifactory's side, and this isn't relevant for the upload command.

eyalbe4 (changes requested):

PR #1277, opened 2021-11-10
Add vcs-fallback feature for Go

yahavi:
-----

Depends on https://github.com/jfrog/jfrog-cli-core/pull/248

If a package does not exist in Artifactory, download it from VCS.
Add the `--vcs-fallback` boolean flag with default true.


go.mod
```diff
@@ -22,7 +22,7 @@ require (
...
 )
 
-
-replace github.com/jfrog/jfrog-client-go => github.com/jfrog/jfrog-client-go dev
- 
-replace github.com/jfrog/gocmd => github.com/jfrog/gocmd master
+replace (
+	github.com/jfrog/gocmd => github.com/yahavi/gocmd v0.1.13-0.20211109135307-6b3e0aca9965
+	github.com/jfrog/jfrog-cli-core/v2 => github.com/yahavi/jfrog-cli-core/v2 v2.0.0-20211110105913-e8e5b8e03bfc
+)
```
yahavi, line 28: Reminder - revert before merge

utils/cliutils/commandsflags.go
```diff
@@ -953,6 +956,10 @@ var flagsMap = map[string]cli.Flag{
 		Name:  nugetV2,
 		Usage: "[Default: false] Set to true if you'd like to use the NuGet V2 protocol when restoring packages from Artifactory.` `",
 	},
+	vcsFallback: cli.BoolTFlag{
+		Name:  vcsFallback,
+		Usage: "[Default: true] Set to false if you don't like to download Go packages from the VCS in case they are missing in Artifactory.` `",
```
eyalbe4, line 961: I suggest we rename the flag to ```no-fallback``` with a ```false``` default value. This means that by default, we will fallback. I think this provides the best user experience. People who are concerned about the implications of this, can always add the flag.
So let's change the flag description to - 
```
[Default: false] Set to true if you'd like to download Go packages from the VCS, in case they are missing in Artifactory
``` 
Let's remember to also modify the code that reads the flag with the new default value, as well as the test.

eyalbe4 (changes requested):

PR #1283, opened 2021-11-14
Shorten executable name to jf

RobiNino:


yahavi (changes requested):
**This is not a code review**

Please also fix the autocompletion scripts under `/completion`. 
For example in Bash:
```bash
complete -F _jfrog -o default jfrog
```
Should be changed to:
```bash
complete -F _jf -o default jf
```
As well as `_jfrog()` => `_jf()`.
`jfrog_bash_completion` =>  `jf_bash_completion`.

The same idea for Zsh.
The script name in generate_scripts.go should also be renamed to `jf`.

Thanks!

eyalbe4 (changes requested):

Issue #1306, opened 2021-11-23
"panic: runtime error: invalid memory address or nil pointer dereference" when running "jfrog rt npmp"


SomaSharath:
**Describe the bug**
When running **jfrog rt npmp --build-name={} --build-number={}** , I get the following error:

`[Info] Running npm Publish
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xb9f71e]

goroutine 1 [running]:
github.com/jfrog/jfrog-cli-core/v2/artifactory/commands/npm.(*NpmPublishCommand).preparePrerequisites(0xc00023d600)
	/root/go/pkg/mod/github.com/jfrog/jfrog-cli-core/v2@v2.5.0/artifactory/commands/npm/publish.go:192 +0xfe
github.com/jfrog/jfrog-cli-core/v2/artifactory/commands/npm.(*NpmPublishCommand).Run(0xc00023d600)
	/root/go/pkg/mod/github.com/jfrog/jfrog-cli-core/v2@v2.5.0/artifactory/commands/npm/publish.go:138 +0x77
github.com/jfrog/jfrog-cli-core/v2/common/commands.Exec({0xf192e0, 0xc00023d600})
	/root/go/pkg/mod/github.com/jfrog/jfrog-cli-core/v2@v2.5.0/common/commands/command.go:26 +0xad
github.com/jfrog/jfrog-cli/artifactory.npmPublishCmd(0xc0000d2f20)
	/var/jenkins_home/workspace/eco-system/release/jfrog-cli-release/temp/jfrog-cli/artifactory/cli.go:1197 +0x1cf
github.com/jfrog/jfrog-cli/artifactory.GetCommands.func33(0x1)
	/var/jenkins_home/workspace/eco-system/release/jfrog-cli-release/temp/jfrog-cli/artifactory/cli.go:537 +0x19
github.com/codegangsta/cli.HandleAction({0xc8cdc0, 0xe06e28}, 0xb)
	/root/go/pkg/mod/github.com/codegangsta/cli@v1.20.0/app.go:490 +0x5a
github.com/codegangsta/cli.Command.Run({{0xdbcd6a, 0xb}, {0x0, 0x0}, {0xc00023a770, 0x1, 0x1}, {0x0, 0x0}, {0x0, ...}, ...}, ...)
[...]

PR #1307, opened 2021-11-24
Fix npm publish NPE

RobiNino:
-----
Following #1306

### DIFF
--- a/npm_test.go
+++ b/npm_test.go
@@ -31,7 +31,8 @@ import (
 
 type npmTestParams struct {
 	testName       string
-	command        string
+	nativeCommand  string
+	legacyCommand  string
 	repo           string
 	npmArgs        string
 	wd             string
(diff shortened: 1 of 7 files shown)


utils/cliutils/utils.go
```diff
@@ -516,30 +516,37 @@ func CreateConfigCmd(c *cli.Context, confType artifactoryUtils.ProjectType) erro
 
 func RunNativeCmdWithDeprecationWarning(cmdName string, projectType artifactoryUtils.ProjectType, c *cli.Context, cmd func(c *cli.Context) error) error {
```
eyalbe4, line 517: For consistency ad readability reasons, I suggest we make this function more similar to the ```RunConfigCmdWithDeprecationWarning``` function, in the sense that the log warning message is placed in a different function.

npm_test.go
```diff
@@ -31,7 +31,8 @@ import (
 
 type npmTestParams struct {
 	testName       string
-	command        string
+	nativeCommand  string
+	legacyCommand  string
```
eyalbe4, line 35: Let's add ```// Deprecated``` comment above this lne, so that we remember to remove it in the future. 
The same goes for the ```TestNpmLegacy``` function.

pip_test.go
```diff
@@ -24,7 +24,15 @@ type PipCmd struct {
 	Options []string
 }
 
-func TestPipInstall(t *testing.T) {
+func TestPipInstallNativeSyntax(t *testing.T) {
+	testPipInstall(t, false)
+}
+
+func TestPipInstallLegacy(t *testing.T) {
```
eyalbe4, line 31: Let's add ```// Deprecated``` comment for this function, so that we remember to remove it in the future.

eyalbe4 (changes requested):

PR #1367, opened 2022-01-03
Add autocorrect to cli cmds

Or-Geva:
-----
If JFrog-CLI detects command typos and can identify exactly one or more valid commands similar to the error, it will try to suggest the correct command(s).
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/jfrog/jfrog-cli
 go 1.14
 
 require (
+	github.com/agnivade/levenshtein v1.1.1
 	github.com/buger/jsonparser v1.1.1
 	github.com/frankban/quicktest v1.13.1 // indirect
 	github.com/go-git/go-git/v5 v5.4.2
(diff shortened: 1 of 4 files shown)


eyalbe4 (changes requested):
The current auto-compete output is as follows - 
```
$ ./jf aa
'aa' is not a jf command. See --help
The most similar commands are
	ag,am,an,ap
```
I suggest modifying it t0 be -
```
$ ./jf aa
'aa' is not a jf command. See --help
The most similar commands are
    jf ag
    jf am
    jf an
    jf ap
```

Or-Geva:
> While the above can probably be fixed by modifying the output layout, the following issue may require some change in the functionality level. See the following example -
> 
> ```
> $ ./jf rt up
> 'up' is not a jf command. See --help
> The most similar commands are
> 	bp,cp,dp,gp,p,pp,sp,u,uc
> ```
> 
> The most similar command however should be - `jf rt u` or `jf rt upload`. I'd be happy to test this further and also review the code once the above is fixed.

We can't know for 100% what the user means by `jf rt up`, maybe `jf rt upload` could be the right answer for some users and `jf rt uc` for others. In my opinion, taking this to the extreme misses the purpose.

eyalbe4 (changes requested):
@Or-Geva,
Following your reply, the only comment left from my side is about the output layoout.

PR #1373, opened 2022-01-11
Remove test.artifactoryProject flag

Or-Geva:
-----
Remove `test.artifactoryProject` that was used by running the tests with a free-tier instance.
This flow is not used anymore since the test.artifactoryProject has a data transfer limit.

### DIFF
--- a/artifactory_test.go
+++ b/artifactory_test.go
@@ -7,8 +7,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


Or-Geva:
The project's tests required an admin access token. GitHub-Actions spins up an Artifactory instance and can't generate this kind of token from the API, so after all, we need project tests separation to work with the Ecosys server.

PR #1430, opened 2022-02-07
Docker scan progress bar

sverdlov93:
-----
Add progress bar for docker scan 

https://user-images.githubusercontent.com/29822394/152808655-f04839ae-3e4f-403a-92cf-1e4d434e35f2.mov


eyalbe4 (changes requested):
This isn't related directly to the code changes in this PR, but this is something we better fix now, because we're relying on this functionality more following this PR. 
The issue is with this function - 
```
// Closes the log file and resets to the default logger
func CloseLogFile(logFile *os.File) {
	if logFile != nil {
		log.SetDefaultLogger()
		err := logFile.Close()
		utils.CheckErrorWithMessage(err, "Failed closing the log file")
	}
}
```
The error returned by the ```utils.CheckErrorWithMessage``` should be returned by this function. We should then see that it is returned upstream, including inside ```defer``` clauses.

PR #1470, opened 2022-03-09
Make all config commands - atomic

sverdlov93:
-----
following https://github.com/jfrog/jfrog-cli-core/pull/352 and https://github.com/jfrog/jfrog-cli-core/issues/350


artifactory_test.go
```diff
@@ -4826,6 +4826,35 @@ func TestConfigAddOverwrite(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestConfigConcurrency(t *testing.T) {
```
eyalbe4, line 4829: I don't think concurrency tests should be included in the jfrog-cli code base. The tests should be deterministic. Having such tests running automatically may produce unexpected results. Let's remove this one.
So besides this test, which is current commented out, I'm actually not sure what this PR adds.

eyalbe4 (changes requested):

PR #1497, opened 2022-04-04
Automatically unlabel 'safe to test' after tests run

talarian1:
-----


.github/workflows/tests.yml
```diff
@@ -14,14 +14,19 @@ jobs:
     if: contains(github.event.pull_request.labels.*.name, 'safe to test') || github.event_name == 'push'
     strategy:
       fail-fast: false
```
yahavi, line 16: This not needed if you're using only 1 agent
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -183,6 +190,7 @@ jobs:
         run: go test -v -timeout 0 --test.docker=true --jfrog.url=${{ secrets.PLATFORM_URL }} --jfrog.adminToken=${{ secrets.PLATFORM_ADMIN_TOKEN }} --rt.dockerRepoDomain=${{ secrets.CONTAINER_REGISTRY }} --rt.dockerVirtualRepo=${{ secrets.DOCKER_VIRTUAL }} --rt.dockerLocalRepo=${{ secrets.DOCKER_LOCAL }} --rt.dockerRemoteRepo=${{ secrets.DOCKER_REMOTE }} --rt.dockerPromoteLocalRepo=${{ secrets.DOCKER_PROMOTE_LOCAL }}
 
   Scripts-tests:
+    needs: Go-Lint
```
yahavi, line 193: This job can safely run without any restrictions
```suggestion
```

yahavi (changes requested):
Let's also rename Go-Lint to something else. I suggest `Pretest`

eyalbe4 (changes requested):
I noticed the following files were removed.
I assume this is because they are unused. Just making sure.
image

talarian1:
@eyalbe4, worse than that. 
Those files, cause the flowing messages to appear :

> 08:28:49 [Warn] The file /Users/yahavi/.nuget/packages/system.reflection.metadata/1.6.0/system.reflection.metadata.1.6.0.nupkg doesn't exist in the NuGet cache directory but it does exist as a target in the assets files. Skipping adding this dependency to the build info. This might be because the package already exists in a different NuGet cache, possibly the SDK's NuGetFallbackFolder cache. Removing the package from this cache may resolve the issue. 

And fails the `TestXrayAuditNugetMultiProject`

PR #1517, opened 2022-04-17
Update the JFrog CLI Plugins Developer Guide

eyalbe4:
-----
* Update the guide to include information about using external resources
* Require go 1.17

### DIFF
--- a/guides/jfrog-cli-plugins-developer-guide.md
+++ b/guides/jfrog-cli-plugins-developer-guide.md
@@ -13,7 +13,7 @@ This article guides you through the process of creating and publishing your own 
 
 ## Creating your first plugin
 ### Prepare your development machine
-* Make sure Go 1.14 or above is installed on your local machine and is included in your system PATH.
+* Make sure Go 1.17 or above is installed on your local machine and is included in your system PATH.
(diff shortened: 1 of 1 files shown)


guides/jfrog-cli-plugins-developer-guide.md
```diff
@@ -99,3 +100,39 @@ To create your own private plugins registry, follow these steps.
...
+```
+Once the plugin's executable is there, you'll be able to see it is installed by just running ```jf```.
+
+## Having your plugin use external resources
+### General
+In some cases your plugin may need to use external resources. For example, the plugin code may need to run an executable or read from a configuration file. You would therefore want these resources to be packaged together with the plugin, so that when it is installed, these resources are also downloaded and become available for the plugin.
+
+### How can you add resources to your plugin?
+The way to include resources for your plugin, is to simply place them inside a directory named `resources` at the root of the plugin's sources directory. You can create any directory structure inside `resources`.
+When publishing the plugin, the content of the `resources` directory is published alongside the plugin executable. When installing the plugin, the resources are also downloaded.
```
gailazar300, line 122: maybe we can add one sentence about executables for different OS
"pay attention to include executables for all plugin's relevant os in your resources directory" - something like this?

PR #1528, opened 2022-05-01
Setup command for Docker Desktop extension

asafgabai:
-----
Added format property to the setup command. Running 'jf setup' with --format=machine prints to stdout when the user finishes filling the registration form and his environment is being built.


general/envsetup/envsetup.go
```diff
@@ -12,12 +9,7 @@ const (
 	registrationPageURL = "https://jfrog.com/start-free/cli/"
 )
 
-func RunEnvSetupCmd() error {
-	fmt.Println()
-	fmt.Println()
-	fmt.Println(coreutils.PrintTitle("Thank you for installing JFrog CLI! 🐸"))
-	fmt.Println(coreutils.PrintTitle("We'll now set up a FREE JFrog environment in the cloud for you, and configure your local machine to use it."))
-	fmt.Println("Your environment will be ready in less than a minute.")
-	setupCmd := envsetup.NewEnvSetupCommand(registrationPageURL)
```
eyalbe4, line 21: The above line should not have been removed. They were moved to *jfrog-cli* from *jfrog-cli-core* recenetly. See my corresponding comment on your dependent *jfrog-cli-core* PR.

utils/cliutils/commandsflags.go
```diff
@@ -446,6 +449,9 @@ const (
 	// *** Completion Commands' flags ***
 	Completion = "completion"
 	Install    = "install"
+
+	// Setup flags
+	setupFormat = "setup-format"
```
eyalbe4, line 454: setupFormat = "setup-format"
-->
outputFormat = "output-format"

utils/cliutils/commandsflags.go
```diff
@@ -1290,6 +1296,10 @@ var flagsMap = map[string]cli.Flag{
 		Name:  Install,
 		Usage: "[Default: false] Set to true to install the completion script instead of printing it to the standard output. ` `",
 	},
+	setupFormat: cli.StringFlag{
+		Name:   "format",
+		Hidden: true,
```
eyalbe4, line 1301: I don't think we need to hide this flag. There's no harm in having it visible.

eyalbe4 (changes requested):

utils/cliutils/commandsflags.go
asafgabai: I chose a specific name for the --format option of the setup command because it's hidden, which makes it different from the --format option in other commands.

PR #1549, opened 2022-05-25
Fix: Terraform module layout in Artifactory

gailazar300:
-----
Change terraform layout in artifactory when uploading module
from : repoName/namespace/provider/moduleName/tag.zip
to: repoName/namespace/moduleName/provider/tag.zip

In addition - fixing terraform's tests by adding a retry mechanism for modules validation.


artifactory_test.go
```diff
@@ -5317,7 +5317,9 @@ func TestTerraformPublish(t *testing.T) {
 	// Download modules to 'result' directory.
 	chdirCallback()
 	assert.NoError(t, os.MkdirAll(tests.Out+"/results/", 0777))
-	runRt(t, "download", tests.TerraformRepo+"/namespace/provider/*", tests.Out+"/results/", "--explode=true")
+	// Wait for "module.json" file to be created in artifactory.
+	time.Sleep(1000 * time.Millisecond)
```
eyalbe4, line 5321: I think it would be better to implement a retry mechanism here, so to avoid timing issues. This is because you can't really tell how much it'll take fot Artifactory to do this. I think we already have tests that do this, and use our retry component.

eyalbe4 (changes requested):

eyalbe4 (approved):

PR #1632, opened 2022-08-01
Add transfer integration tests

yahavi:
-----

Depends on https://github.com/jfrog/jfrog-cli-core/pull/479
See running here: https://github.com/yahavi/jfrog-cli/actions/runs/2775810147

Add transfer suite that runs `transfer-files` between local on-prem Artifactory and a SaaS server:
* TestTransferTwoRepos - test transfer of 2 repositories
* TestTransferExcludeRepo - test exclusion of 1 repository
* TestTransferMaven - test transfer of Maven repository and make sure the metadata file is created in the target
* TestTransferPaginationAndThreads - test the AQL pagination code and change the number of threads on runtime

Usage:
```
go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.transfer --jfrog.targetUrl=<platform-url> --jfrog.targetAdminToken=<target-access-token> --jfrog.home=<local-artifactory-home>
```


inttestutils/transfer.go
```diff
@@ -0,0 +1,240 @@
...
+// serverDetails - Source Artifactory server details
+func RefreshStorageInfoAndWait(sourceServerDetails *config.ServerDetails) error {
+	// Create service manager
+	serviceManager, err := utils.CreateServiceManager(sourceServerDetails, -1, 0, false)
+	if err != nil {
+		return err
+	}
+
+	// Refresh and get storage info with refresh=true.
+	storageInfo, err := serviceManager.StorageInfo(true)
```
RobiNino, line 119: Since the refresh is async, this makes this REST require the following functionality, which makes it pretty useless. WDYT? We should maybe split it to two separate apis.

inttestutils/transfer.go
```diff
@@ -0,0 +1,240 @@
...
+// pattern - Search wildcard pattern
+// serverDetails - The Artifactory server details
+// t - The testing object
+func CountArtifactsInPath(pattern string, serverDetails *config.ServerDetails, t *testing.T) int {
+	searchCmd := generic.NewSearchCommand()
+	searchCmd.SetServerDetails(serverDetails).SetSpec(spec.NewBuilder().Pattern(pattern).BuildSpec())
+	reader, err := searchCmd.Search()
+	assert.NoError(t, err)
+	defer assert.NoError(t, reader.Close())
+	lenght, err := reader.Length()
```
RobiNino, line 164: typo

inttestutils/transfer.go
```diff
@@ -0,0 +1,240 @@
...
+				time.Sleep(time.Second)
+				continue
+			}
+			conf := &utils.TransferSettings{ThreadsNumber: threadsCount + 1}
+			assert.NoError(t, utils.SaveTransferSettings(conf))
+			break
+		}
+
+		// Wait for the number of threads to be increase by 1
+		threadCountChanged := false
```
RobiNino, line 228: I think this variable is redundant. We can just return when the threads count change (assert correct number), and `assert.Fail(t, "number of threads did not change as expected...")` at the end of the function.

inttestutils/transfer.go
```diff
@@ -0,0 +1,240 @@
...
+const (
+	dataTransferUrl = "https://releases.jfrog.io/artifactory/jfrog-releases/data-transfer/[RELEASE]/"
+	groovyFileName  = "dataTransfer.groovy"
+	jarFileName     = "data-transfer.jar"
+)
+
+// Create test repositories in the target Artifactory
+// targetArtifactoryCli - Target Artifactory CLI
+func CreateTargetRepos(targetArtifactoryCli *tests.JfrogCli) {
+	log.Info("Creating repositiries in target Artifactory...")
```
RobiNino, line 34:
```suggestion
	log.Info("Creating repositories in target Artifactory...")
```

transfer_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func initTransferTest(t *testing.T) func() {
+	if !*tests.TestTransfer {
+		t.Skip("Skipping transfer test. To run transfer test add the '-test.transfer=true' option.")
+	}
+	oldHomeDir, newHomeDir := prepareHomeDir(t)
+
+	// Delete the target server if exist
+	config, err := commands.GetConfig("target", false)
```
RobiNino, line 48: I suggest we give this server-id a uniquer name (`jfrog-cli-test-target` for example), and make it a constant.

inttestutils/transfer.go
```diff
@@ -0,0 +1,240 @@
...
+		threadCountChanged := false
+		for !threadCountChanged && !*filesTransferFinished {
+			if transferfiles.GetThreads() == threadsCount+1 {
+				threadCountChanged = true
+				break
+			}
+			time.Sleep(time.Second)
+		}
+
+		// If false, the transfer-files process is finished before the threads changed
```
RobiNino, line 237: This could be a race. We don't have to fix it now, but we might run into it once the 15 idle seconds are reduced.

transfer_test.go
```diff
@@ -0,0 +1,161 @@
...
+	assert.NoError(t, err)
+	verifyExistInArtifactory(tests.GetTransferExpectedRepo1(), repo1Spec, t)
+	repo2Spec, err := tests.CreateSpec(tests.SearchRepo2)
+	assert.NoError(t, err)
+	verifyExistInArtifactory(tests.GetTransferExpectedRepo2(), repo2Spec, t)
+
+	// Execute transfer-files
+	assert.NoError(t, artifactoryCli.WithoutCredentials().Exec("transfer-files", "default", "target", "--include-repos="+tests.RtRepo1+";"+tests.RtRepo2))
+
+	// Verify files were transferred to the target Artifactory
```
RobiNino, line 86: I suggest verifying they also still exist in source, just in case.

transfer_test.go
```diff
@@ -0,0 +1,161 @@
...
+	inttestutils.VerifyExistInTargetArtifactory(tests.GetTransferExpectedRepo1(), repo1Spec, targetServerDetails, t)
+	inttestutils.VerifyExistInTargetArtifactory(tests.GetTransferExpectedRepo2(), repo2Spec, targetServerDetails, t)
+}
+
+func TestTransferExcludeRepo(t *testing.T) {
```
RobiNino, line 91: can possibly use a common function with `TestTransferTwoRepos`.

transfer_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Execute transfer-files
+	assert.NoError(t, artifactoryCli.WithoutCredentials().Exec("transfer-files", "default", "target", "--include-repos="+tests.RtRepo1+";"+tests.RtRepo2, "--exclude-repos="+tests.RtRepo2))
+
+	// Verify repo1 files were transferred to the target Artifactory
+	inttestutils.VerifyExistInTargetArtifactory(tests.GetTransferExpectedRepo1(), repo1Spec, targetServerDetails, t)
+	inttestutils.VerifyExistInTargetArtifactory([]string{}, repo2Spec, targetServerDetails, t)
+}
+
+func TestTransferMaven(t *testing.T) {
```
RobiNino, line 115: An improvement to this test would be to assert the delayed files do not exists in target until handled. Not sure if this is easily feasible.

transfer_test.go
```diff
@@ -0,0 +1,161 @@
...
+	defer cleanUp()
+
+	// Decrease the AQL pagination limit to 10
+	transferfiles.AqlPaginationLimit = 10
+	defer func() {
+		transferfiles.AqlPaginationLimit = transferfiles.DefaultAqlPaginationLimit
+	}()
+
+	// Upload 100 files to the source Artifactory
+	for i := 0; i < 100; i++ {
```
RobiNino, line 147: You may want to change this number to a number that is divided with the limit with a remainder (101)

transfer_test.go
```diff
@@ -0,0 +1,161 @@
...
+		fileIndexStr := strconv.FormatInt(int64(i), 10)
+		assert.NoError(t, artifactoryCli.WithoutCredentials().Exec("curl", "-s", "-XPUT", tests.RtRepo1+"/"+fileIndexStr, "-d "+fileIndexStr))
+	}
+
+	// Asynchronously exec transfer-files and increse the threads count by 1
```
RobiNino, line 152:
```suggestion
	// Asynchronously exec transfer-files and increase the threads count by 1
```

RobiNino (changes requested):
Awesome thanks!

inttestutils/transfer.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/jfrog/jfrog-client-go/utils/log"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	dataTransferUrl = "https://releases.jfrog.io/artifactory/jfrog-releases/data-transfer/[RELEASE]/"
+	groovyFileName  = "dataTransfer.groovy"
+	jarFileName     = "data-transfer.jar"
+)
+
```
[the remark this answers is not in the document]
yahavi: I made the change to create/clean/delete repositories from the pre-configured tests repositories.
I'm using the existing test creates/deletes in the source Artifactory, but this is a case for the target Artifactory.

inttestutils/transfer.go
yahavi: It is a race, but a race between a rabbit and a turtle. This code runs while uploading 101 files. If needed in the future, we can change 101 to a bigger number, although I'm skeptical.

transfer_test.go
yahavi: I don't think it is possible, because Artifactory can index the Maven repository at the second we upload the pom.xml.

RobiNino: I wasn't referring to the metadata files, but to the delayed files (pom.xml itself)

inttestutils/transfer.go
yahavi: I agree. We should talk about it and decide how to make it right.

Issue #1636, opened 2022-08-05
Skip checksum verification be chosen when downloading & uploading


mikesun666:
**Is your feature request related to a problem? Please describe.**
JFrog CLI download & upload can skip checksum verification

**Describe the solution you'd like to see**
JFrog CLI uses the checksum deploy is mandatory right now. But when the customer can ensure there is no duplicate file in Artifactory, the checksum deploy won't be necessary. 
In addition, When JFrog CLI download a big file, the checksum verification will be slow, and the customer wants they can skip it sometimes. 
So the checksum verification that can be chosen is better. 

Like the image below, It looks like the download is finished, but the file size is too big, JFrog CLI doing the checksum verification spent a lot of time.
![image](https://user-images.githubusercontent.com/43203674/182995460-24e878b1-c50c-4475-a73c-1a91a3693df1.png)

mikesun666:
@SteveNewson I find we have already supported the download checksum skip, so do we consider supporting upload skip as well?

PR #1661, opened 2022-08-17
Add ability to ignore checksum failures when downloading files

SteveNewson:
-----
Filtered resources manipulate the contents of artifacts during download.
The SHA1 supplied by the server does not reflect those changes, but
instead only hashes the raw unfiltered template. In the current
implementation, the client always checks that the hash matches the
expected hash and returns an error if they do not match.

This change adds a flag to indicate whether the checksum should be
validated defaulting to false.
```
jfrog rt dl repo/blah.filtered --skip-checksum
```
Feature request: https://github.com/jfrog/jfrog-cli/issues/1636

### DIFF
--- a/artifactory_test.go
+++ b/artifactory_test.go
@@ -207,6 +207,27 @@ func TestArtifactoryUploadPathWithSpecialCharsAsNoRegex(t *testing.T) {
 	cleanArtifactoryTest()
 }
 
+func TestArtifactoryIgnoreChecksumOnFilteredResource(t *testing.T) {
+	initArtifactoryTest(t, "")
+	filePath := "testdata/filtered/a.txt"
+
+	// Upload a filtered resource
+	runRt(t, "upload", filePath, tests.RtRepo1, "--flat", "--target-props", "artifactory.filtered=true")
+	searchFilePath, err := tests.CreateSpec(tests.SearchAllRepo1)
+	assert.NoError(t, err)
+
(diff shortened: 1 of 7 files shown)


artifactory_test.go
```diff
@@ -207,6 +207,27 @@ func TestArtifactoryUploadPathWithSpecialCharsAsNoRegex(t *testing.T) {
...
+	initArtifactoryTest(t, "")
+	filePath := "testdata/filtered/a.txt"
+
+	// Upload a filtered resource
+	runRt(t, "upload", filePath, tests.RtRepo1, "--flat", "--target-props", "artifactory.filtered=true")
+	searchFilePath, err := tests.CreateSpec(tests.SearchAllRepo1)
+	assert.NoError(t, err)
+
+	inttestutils.VerifyExistInArtifactory(tests.GetSimpleUploadFilteredRepo1(), searchFilePath, serverDetails, t)
+
```
yahavi, line 220: Let's disable logging to prevent "Error"s in the logs
```suggestion
	// Discard output logging to prevent negative logs
	previousLogger := tests.RedirectLogOutputToNil()
	defer log.SetLogger(previousLogger)
```

yahavi (changes requested):
Thanks again, @SteveNewson!
The added test is brilliant! 🥇

yahavi (approved):

PR #1664, opened 2022-08-24
Add JFROG_CLI_FAIL_NO_OP environment variable

Or-Geva:
-----
This PR adds a new environment variable 'JFROG_CLI_FAIL_NO_OP' is added to allow fail-no-op. You can now set it by flag, which is the first priority, or by the environment, which is the second priority.

Depends on: https://github.com/jfrog/jfrog-cli-core/pull/519


eyalbe4 (changes requested):
Having the following code duplicated multiple times isn't a good practice. Let's redesign the code to avoid this duplication. 
I'd be happy to review this PR again after this change.
```
const EnvVar string = `	JFROG_CLI_FAIL_NO_OP
	[Default: false]
	Set to true if you'd like the command to return exit code 2 in case of no files are affected.`
```

PR #1672, opened 2022-08-28
Add installDataTransferPlugin to transfer tests

yahavi:
-----

Add new `-test.installDataTransferPlugin` flag to the transfer tests.
The reason for this change is to allow the data-transfer plugin tests to install the latest data-transfer code instead of the latest release.

### DIFF
--- a/.github/workflows/transferTests.yml
+++ b/.github/workflows/transferTests.yml
@@ -46,4 +46,4 @@ jobs:
           JFROG_HOME: ${{ runner.temp }}
 
       - name: Run transfer tests
-        run: go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.transfer --jfrog.targetUrl=${{ secrets.PLATFORM_URL }} --jfrog.targetAdminToken=${{ secrets.PLATFORM_ADMIN_TOKEN }} --jfrog.home=${{ runner.temp }} --ci.runId=${{ runner.os }}-transfer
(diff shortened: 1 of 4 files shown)


.github/workflows/transferTests.yml
```diff
@@ -46,4 +46,4 @@ jobs:
           JFROG_HOME: ${{ runner.temp }}
 
       - name: Run transfer tests
-        run: go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.transfer --jfrog.targetUrl=${{ secrets.PLATFORM_URL }} --jfrog.targetAdminToken=${{ secrets.PLATFORM_ADMIN_TOKEN }} --jfrog.home=${{ runner.temp }} --ci.runId=${{ runner.os }}-transfer
+        run: go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.transfer --test.installDataTransferPlugin --jfrog.targetUrl=${{ secrets.PLATFORM_URL }} --jfrog.targetAdminToken=${{ secrets.PLATFORM_ADMIN_TOKEN }} --jfrog.home=${{ runner.temp }} --ci.runId=${{ runner.os }}-transfer
```
eyalbe4, line 49: Should the flag namne be `reloadDataTransferPlugin` instead of `installDataTransferPlugin`?
Because all this flag is runs the "reload plugins" REST API and doesn't really place the plugin files on the server. Let's also change the descriptions in the README and in -
```
InstallDataTransferPlugin = flag.Bool("test.installDataTransferPlugin", false, "Install data-transfer plugin on the source Artifactory server")
```
accordingly.

Which brings me to the next questions - why can't we always reload the plugins as part of the transfer tests, and remove this flag altogether? Why do we need a flag for it? Is there a scenario where reloading the plugin is harmful?

eyalbe4 (changes requested):

.github/workflows/transferTests.yml
yahavi: That is not accurate - this flag prevents installing and reloading the Data Transfer plugin.

eyalbe4: You're right. NP.

eyalbe4 (approved):

PR #1681, opened 2022-09-04
Sync-deletes - The wrong folders are deleted when uploading files with uppercase names

Or-Geva:
-----
Part of: https://github.com/jfrog/jfrog-client-go/pull/646

### DIFF
--- a/artifactory_test.go
+++ b/artifactory_test.go
@@ -995,6 +995,7 @@ func TestArtifactoryUploadAndExplode(t *testing.T) {
 }
 
 func TestUploadAndSyncDeleteCaseSensitivity(t *testing.T) {
+	initArtifactoryTest(t, "")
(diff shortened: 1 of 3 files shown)


artifactory_test.go
```diff
@@ -994,6 +994,17 @@ func TestArtifactoryUploadAndExplode(t *testing.T) {
 	cleanArtifactoryTest()
 }
 
+func TestUploadAndSyncDeleteCaseSensitivity(t *testing.T) {
+	initArtifactoryTest(t, "")
+	runRt(t, "upload", "--sync-deletes",tests.RtRepo1+"/", path.Join("testdata", "syncdeletes", "*"), tests.RtRepo1+"/")
```
talarian1, line 999: Would it be better to make sure files are being deleted by doing two separate uploads?

PR #1744, opened 2022-10-30
Add install data transfer plugin command

attiasas:
-----


docs/artifactory/installtransfer/help.go
```diff
@@ -0,0 +1,12 @@
...
+
+var Usage = []string{"rt transfer-install <server-id> [command options]"}
+
+func GetDescription() string {
+	return "Download and install the data-transfer user plugin."
+}
+
+func GetArguments() string {
+	return `	server-id
+		The source server ID. The plugin will be install on the primary node of this server.`
```
yahavi, line 11:
```suggestion
		The source server ID. The plugin will be installed on the primary node of this server.`
```
Also - It is not clear if `The plugin will be installed on the primary node of this server` is a description or a precondition. If it is a precondition, it shouldn't be in this flag, but in the command description.

docs/artifactory/installtransfer/help.go
```diff
@@ -0,0 +1,12 @@
+package installtransfer
+
+var Usage = []string{"rt transfer-install <server-id> [command options]"}
+
+func GetDescription() string {
+	return "Download and install the data-transfer user plugin."
```
yahavi, line 6: Let's add that the user must run this command in the source Artifactory machine

utils/cliutils/commandsflags.go
```diff
@@ -1403,6 +1412,18 @@ var flagsMap = map[string]cli.Flag{
 		Name:  Status,
 		Usage: "[Default: false] Set to true to show the status of the transfer-files command currently in progress.` `",
 	},
+	InstallPluginVersion: cli.StringFlag{
+		Name:  version,
+		Usage: "[Optional] The plugin version to download and install.` `",
```
yahavi, line 1417:
```suggestion
		Usage: "[Optional, default: latest] The data-transfer plugin version to download and install.` `",
```

utils/cliutils/commandsflags.go
```diff
@@ -1403,6 +1412,18 @@ var flagsMap = map[string]cli.Flag{
...
+	},
+	InstallPluginSrcDir: cli.StringFlag{
+		Name:  InstallPluginSrcDir,
+		Usage: "[Optional] The local directory that contains the plugin files to install.` `",
```
yahavi, line 1421: Let's add "Use this option to ..."
We should also explain what is the content of this directory.

utils/cliutils/commandsflags.go
```diff
@@ -1403,6 +1412,18 @@ var flagsMap = map[string]cli.Flag{
...
+	},
+	InstallPluginHomeDir: cli.StringFlag{
+		Name:  InstallPluginHomeDir,
+		Usage: "[Optional] The local Jfrog home directory to install the plugin in.` `",
```
yahavi, line 1425:
```suggestion
		Usage: "[Optional, default: /opt/jfrog] The local JFrog home directory to install the plugin in.` `",
```

artifactory/cli.go
```diff
@@ -2326,6 +2341,33 @@ func transferConfigCmd(c *cli.Context) error {
 	return nil
 }
 
+func transferInstallCmd(c *cli.Context) error {
+	if c.NArg() != 1 {
+		return cliutils.WrongNumberOfArgumentsHandler(c)
+	}
+	// Get the artifactory server
+	serverDetails, err := coreConfig.GetSpecificConfig(c.Args()[0], false, true)
```
yahavi, line 2349: Suggestion - print an error and exit if the server URL does not contain `127.0.0.1` or `localhost`

yahavi (changes requested):
Please ask @eyalbe4 for a grammar review.

artifactory/cli.go
```diff
@@ -2326,6 +2341,33 @@ func transferConfigCmd(c *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+	installCmd := transferinstall.NewInstallTransferCommand(serverDetails)
+	// Optional flags
+	versionToInstall := c.String(cliutils.InstallPluginVersion)
```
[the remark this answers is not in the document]
attiasas: done, flag is --version

docs/artifactory/installtransfer/help.go
attiasas: done, moved to description

docs/artifactory/installtransfer/help.go
attiasas: done

utils/cliutils/commandsflags.go
attiasas: done the default part, since I have implemented this as a general install plugin, so I don't think we should add a specific plugin name to the usage of this general flag

utils/cliutils/commandsflags.go
attiasas: done

artifactory/cli.go
attiasas: do we need to constraint it to localhosts only? we already have a check that makes sure we can find the plugin directory in the machine, if it won't find it will exit with an error that suggest that it must run on a machine with Artifactory.

utils/cliutils/commandsflags.go
attiasas: done

artifactory/cli.go
```diff
@@ -2326,6 +2341,32 @@ func transferConfigCmd(c *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+	installCmd := transferinstall.NewInstallDataTransferCommand(serverDetails)
+	// Optional flags
+	versionToInstall := c.String(cliutils.Version)
+	if versionToInstall != "" {
+		installCmd.SetInstallVersion(version.NewVersion(versionToInstall))
+	}
+	sourceFilesPath := c.String(cliutils.InstallPluginSrcDir)
```
eyalbe4, line 2359: I think the command should not accept a version flag with a "path to files" flag. Only one of these flags should be accpetted by the command. See how we perform such input validations for other commands.

eyalbe4 (changes requested):

PR #1746, opened 2022-10-30
Add Frogbot scan-and-fix

yahavi:
-----

### DIFF
--- a/.github/workflows/frogbot-scan-and-fix.yml
+++ b/.github/workflows/frogbot-scan-and-fix.yml
@@ -1,0 +1,36 @@
+name: "Frogbot Scan and Fix"
+on:
+  push:
+    # Creating fix pull requests will be triggered by any push to one of the these branches.
+    # You can add or replace to any branch you want to open fix pull requests for.
+    branches:
+      - "dev"
+permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/frogbot-scan-and-fix.yml
```diff
@@ -0,0 +1,36 @@
+name: "Frogbot Scan and Fix"
+on:
+  push:
+    # Creating fix pull requests will be triggered by any push to one of the these branches.
+    # You can add or replace to any branch you want to open fix pull requests for.
+    branches:
+      - "dev"
```
talarian1, line 7: Why not also include "master/main"?

Or-Geva (approved):

.github/workflows/frogbot-scan-and-fix.yml
yahavi: Because we want to merge fixes only to the dev branch, which are merged to the v2 branch during the release.
We do the same in the Core package - https://github.com/jfrog/jfrog-cli-core/blob/v2.23.2/.github/workflows/frogbot-scan-and-fix.yml#L7.

PR #1845, opened 2023-02-23
Fix node v18+ .nprmc private registry E401 issue

omerzi:
-----


npm_test.go
```diff
@@ -327,6 +327,7 @@ func prepareArtifactoryForNpmBuild(t *testing.T, workingDirectory string) {
 	// Run install with -cache argument to download the artifacts from Artifactory
 	// This done to be sure the artifacts exists in Artifactory
 	jfrogCli := tests.NewJfrogCli(execMain, "jfrog", "")
+	log.Info("running cli with postinstall")
```
yahavi, line 330: I think this would be printed in all cases
```suggestion
```

yahavi (changes requested):
Please don't forget to update go.mod and go.sum before merging the code

PR #1881, opened 2023-03-29
Fix Distribution Tests and delay Windows Poetry Test

sverdlov93:
- [ ] All [static analysis checks](https://github.com/jfrog/jfrog-cli/actions/workflows/analysis.yml) passed.
-----

### DIFF
--- a/utils/tests/utils.go
+++ b/utils/tests/utils.go
@@ -774,6 +774,6 @@ func SkipKnownFailingTestIfPossible(t *testing.T) {
 	if time.Now().Month() < skipMonth {
 		t.Skip("Skipping a known failing test, will resume testing after ", skipMonth.String())
 	} else {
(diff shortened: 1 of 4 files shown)


utils/tests/utils.go
```diff
@@ -768,3 +768,12 @@ func VerifySha256DetailedSummaryFromResult(t *testing.T, result *commandutils.Re
 		}
 	}
 }
+
+func SkipKnownFailingTestIfPossible(t *testing.T) {
+	skipMonth := time.July
```
talarian1, line 773: I think it makes more sense to receive the skipping period from the test that is being skipped as it might have a known fixing date.  
And then the name can be more clear - maybe SkipKnownFailingTestUntil()

talarian1, line 773: It will stop working when we will need to overlap a year.

talarian1 (changes requested):

Issue #1836, opened 2023-02-19
jf docker scan: Provide an option to filter vulnerabilities by severities


harshagv:
**Is your feature request related to a problem? Please describe.**
While performing the scan of any docker (containers) image using Jfrog Xray, we currently don't have the option to filter vulnerabilities by severities

**Describe the solution you'd like to see**
Command line option to option to filter vulnerabilities by severities
```bash
jf docker scan <IMAGE:TAG> --severity="HIGH,CRITICAL"
```

For example: trivy scanner provides an option that meets this requirement
```bash
trivy image --severity HIGH,CRITICAL httpd:latest
```

**Describe alternatives you've considered**
Currently, I am using jq filter to list only the vulnerabilities by severities and it only helps in case of `--format=json|simple-json`. But this would not help in case of  `--format=table|sarif`
e.g: 

```bash
jq -r '.[] | (.vulnerabilities[] | select(.severity == "High" or .severity == "CRITICAL"))
```

**Additional context**

ventz:
+1 on this request!!

Also - please provide a way to filter out things that have no fix (similar to Trivy, Snyk, Orca, etc)

sverdlov93:
Hi @harshagv and @ventz,
The above functionality can be achieved by using JFrog Xray Watches functionality.
Create a Watch on Xray (or use an existing one) and then connect a policy to it according to your requirements:
![Screenshot 2023-03-21 at 16 43 21](https://user-images.githubusercontent.com/29822394/226642297-0e916bf3-cfd6-41bd-8642-2806718b0d23.png)

Then you can run `jf docker scan / jf scan / jf audit` commands with the `--watches` flag: 

![Screenshot 2023-03-21 at 16 44 43](https://user-images.githubusercontent.com/29822394/226642732-0a66bcf0-b6f2-4a17-b67d-8acdfa058bcc.png)

For example: 
`jf docker scan alpine:latest --watches=mywatch`

ventz:
@sverdlov93 Thank you for mentioning this!

Would it be possible to add stand alone client-side filtering -- effectively a json filter similar to @harshagv's jq select, to the binary itself?

The use case being that when this is distributed in an action to groups/teams, it would require for them to pre-create this, rather than just have a "drop in" solution to scan containers.

We are seeing most groups going with Trivy due to the "drop in" container scanning ability within GitHub Actions. The results are pulled/filtered from the json output, and then github issues are created for each.

harshagv:
[...]

PR #1934, opened 2023-05-04
Fix docker scan missing filter flags and wrong image name parsing

sverdlov93:
- [ ] All [static analysis checks](https://github.com/jfrog/jfrog-cli/actions/workflows/analysis.yml) passed.
-----
should fix https://github.com/jfrog/jfrog-cli/issues/1836

### DIFF
--- a/buildtools/cli.go
+++ b/buildtools/cli.go
@@ -717,7 +717,7 @@ func dockerCmd(c *cli.Context) error {
 	default:
 		err = dockerNativeCmd(c)
 	}
-	if err == nil && cmd != "scan" {
+	if err == nil {
 		log.Output(coreutils.PrintTitle("Hint: Use 'jf docker scan' to scan a local Docker image for security vulnerabilities with JFrog Xray"))
 	}
 	return err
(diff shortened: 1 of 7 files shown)


buildtools/cli.go
```diff
@@ -717,7 +717,7 @@ func dockerCmd(c *cli.Context) error {
 	default:
 		err = dockerNativeCmd(c)
 	}
-	if err == nil {
+	if err == nil && cmd != "scan" {
```
omerzi, line 720: How come `cmd == scan` in this case when line 716 returns from this function for this case?

utils/cliutils/codegangstautils.go
```diff
@@ -39,9 +40,7 @@ func GetThreadsCount(c *cli.Context) (threads int, err error) {
 }
 
 func ExtractCommand(c *cli.Context) (command []string) {
```
omerzi, line 42: The named return value is redundant

omerzi (commented):
LGTM

PR #2186, opened 2023-09-05
Run Xray tests with latest analyzer manager version

sverdlov93:

---

### DIFF
--- a/.github/workflows/xrayTests.yml
+++ b/.github/workflows/xrayTests.yml
@@ -23,6 +23,8 @@ jobs:
     runs-on: ${{ matrix.os }}-latest
     env:
       GRADLE_OPTS: -Dorg.gradle.daemon=false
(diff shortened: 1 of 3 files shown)


.github/workflows/xrayTests.yml
```diff
@@ -23,6 +23,8 @@ jobs:
     runs-on: ${{ matrix.os }}-latest
     env:
       GRADLE_OPTS: -Dorg.gradle.daemon=false
+      # Run Xray tests with latest Analyzer
+      JFROG_CLI_ANALYZER_MANAGER_VERSION: "[RELEASE]"
```
attiasas, line 27: If we put this env var for every pull request it will cause failure to all our tests until they will fix the analyzer manager, I would make only the integration tests runs `cron` with this env

PR #2193, opened 2023-09-07
Support Platform Access Token

RobiNino:

---
Note the command currently only supports authentication using an Access Token.
Depends on https://github.com/jfrog/jfrog-cli-core/pull/938


docs/general/token/help.go
```diff
@@ -0,0 +1,14 @@
+package token
+
+var Usage = []string{"atc", "atc <username>"}
+
+func GetDescription() string {
+	return `Creates an access token.
+		By default, an user-scoped token will be created. 
+		Administrator may provide the scope explicitly with '--scope', or implicitly with '--groups', '--grant-admin'`
```
yahavi, line 8:
```suggestion
		Administrator may provide the scope explicitly with '--scope', or implicitly with '--groups', '--grant-admin'.`
```

utils/cliutils/commandsflags.go
```diff
@@ -1141,26 +1173,26 @@ var flagsMap = map[string]cli.Flag{
 		Name:  vars,
 		Usage: "[Optional] List of variables in the form of \"key1=value1;key2=value2;...\" to be replaced in the template. In the template, the variables should be used as follows: ${key1}.` `",
 	},
-	groups: cli.StringFlag{
-		Name: groups,
+	rtAtcGroups: cli.StringFlag{
+		Name: Groups,
 		Usage: "[Default: *] A list of comma-separated groups for the access token to be associated with. " +
 			"Specify * to indicate that this is a 'user-scoped token', i.e., the token provides the same access privileges that the current subject has, and is therefore evaluated dynamically. " +
```
yahavi, line 1179:
```suggestion
			"Specify * to indicate that this is a 'user-scoped token', i.e., the token provides the same access privileges that the current subject has, and is therefore evaluated dynamically.` `" +
```

yahavi (changes requested):

utils/cliutils/commandsflags.go
RobiNino: there's another line to the usage and it has the suffix

artifactory/cli.go
```diff
@@ -2199,7 +2200,7 @@ func groupDeleteCmd(c *cli.Context) error {
 	return commands.Exec(groupDeleteCmd)
 }
 
-func accessTokenCreateCmd(c *cli.Context) error {
+func artifactoryAccessTokenCreateCmd(c *cli.Context) error {
```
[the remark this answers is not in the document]
RobiNino: Since the command currently only supports authentication using access tokens, we rather not send a deprecation notice at the moment.

PR #2283, opened 2023-10-24
Audit SCA recursive scan

attiasas:

---

Depends on https://github.com/jfrog/jfrog-cli-core/pull/994

We are introducing a new parameter to the Sca `Audit` command:

* `exclusions` -  You can specify exclusion patterns (wildcards) for files that should not be included in the mapping when detecting technologies, if not specified it will use the default exclusions set: ```["*node_modules*", "*target*", "*venv*", "*test*"]```

These parameters and capabilities will integrate with existing options as follows:

* `workingDir` -  If it doesn't exist, it will recursively detect the projects (and technologies) inside the current directory. If at least one value exists, it will perform technology detection only on the given directories without recursion.
* Technologies flags (--npm, --maven, etc.) - If a specific technology flag is specified but not detected, it will only scan the root (the given working directory) with that technology.


utils/cliutils/commandsflags.go
```diff
@@ -1380,6 +1381,10 @@ var flagsMap = map[string]cli.Flag{
 		Name:  workingDirs,
 		Usage: "[Optional] A comma-separated list of relative working directories, to determine audit targets locations.` `",
 	},
+	ExclusionsAudit: cli.StringFlag{
+		Name:  exclusions,
+		Usage: "[Default: *node_modules*;*target*;*venv*;*test*] Semicolon-separated list of exclusions. Exclusions can include the * and the ? wildcards.` `",
```
eyalbe4, line 1386: Semicolon-separated list of exclusions. Exclusions can include the * and the ? wildcards.
-->
List of exclusions separated by semicolons, utilized to skip sub-projects from undergoing an audit. These exclusions may incorporate the * and ? wildcards.

eyalbe4 (changes requested):

PR #2323, opened 2023-11-20
Remove poetry integration tests for build-info

attiasas:

---

Poetry is not supported for build-info, only for Audit.
Remove related tests


utils/tests/consts.go
```diff
@@ -176,8 +176,6 @@ var (
 	PypiVirtualRepo        = "cli-pypi-virtual"
```
yahavi, line 176: Let's remove also: 
* PoetryRemoteRepositoryConfig
* PoetryVirtualRepositoryConfig
* PoetryBuildName

yahavi (changes requested):
There are more files to delete under `testdata/poetry`

PR #2393, opened 2023-12-28
Simplify Jenkinsfile

eyalbe4:
This pull request simplifies the JFrog CLI release process.
It does this by removing the EXECUTION_MODE variable from the Jenkinsfile.
With this change, the pipeline should be triggered once, instead of twice, without the need to set the values to the EXECUTION_MODE variable before the rub execution.


yahavi (approved):
LGTM

PR #2472, opened 2024-03-11
Artifactory Release Lifecycle Management - Add Import bundle command

EyalDelarea:

---


docs/lifecycle/importbundle/help.go
```diff
@@ -0,0 +1,13 @@
+package importbundle
+
+var Usage = []string{"rbi [command options] <path to bundle>"}
+
+func GetDescription() string {
+	return "Import a Release Bundle archive"
```
RobiNino, line 6:
```suggestion
	return "Import a local release bundle archive to Artifactory"
```

docs/lifecycle/importbundle/help.go
```diff
@@ -0,0 +1,13 @@
+package importbundle
+
+var Usage = []string{"rbi [command options] <path to bundle>"}
```
RobiNino, line 3:
```suggestion
var Usage = []string{"rbi [command options] <path to archive>"}
```

docs/lifecycle/importbundle/help.go
```diff
@@ -0,0 +1,13 @@
...
+
+func GetDescription() string {
+	return "Import a Release Bundle archive"
+}
+
+func GetArguments() string {
+	return `	path to bundle
+		Path for the desired imported bundle
```
RobiNino, line 11:
```suggestion
	return `	path to archive
		Path to the release bundle archive on the filesystem. 
```

lifecycle/cli.go
```diff
@@ -296,6 +309,22 @@ func export(c *cli.Context) error {
 	return commands.Exec(exportCmd)
 }
 
+func releaseBundleImport(c *cli.Context) error {
```
RobiNino, line 312: Assert that exactly one argument was provided and check if the help is requested.
Missing on the export function too. Something like this:

```
	if show, err := cliutils.ShowCmdHelpIfNeeded(c, c.Args()); show || err != nil {
		return err
	}

	if c.NArg() != 1 {
		return cliutils.WrongNumberOfArgumentsHandler(c)
	}

File: lifecycle_test.go [hunk 5]
```diff
@@ -148,6 +148,9 @@ func TestLifecycleFullFlow(t *testing.T) {
 	exportRb(t, tests.LcRbName2, number2, tempDir)
 	defer deleteExportedReleaseBundle(t, tests.LcRbName2)
 
+	// Test import
+	importRb(t, tests.LcRbName2, number2)
```
RobiNino, line 152: If I'm not mistaken, the release bundle already exists before importing. Consider moving the import (and maybe export as well) to a separate test, to also avoid this test being too long

lifecycle_test.go
```diff
@@ -446,6 +449,19 @@ func sendGpgKeyPair() {
 	coreutils.ExitOnErr(errorutils.CheckResponseStatusWithBody(resp, body, http.StatusCreated))
 }
 
+func importRb(t *testing.T, rbName, rbVersion string) {
+	wd, err := os.Getwd()
+	assert.NoError(t, err)
+	exportedFileName := rbName + "-" + rbVersion + ".zip"
```
RobiNino, line 455: We can also test the scenario where the exported release bundle has a custom name

lifecycle_test.go
```diff
@@ -446,6 +449,19 @@ func sendGpgKeyPair() {
...
+	exportedFilePath := path.Join(wd, rbName, rbVersion, exportedFileName)
```
RobiNino, line 456:
```suggestion
	exportedFilePath := filepath.Join(wd, rbName, rbVersion, exportedFileName)
```
To support windows

lifecycle_test.go
```diff
@@ -446,6 +449,19 @@ func sendGpgKeyPair() {
...
+	output := lcCli.RunCliCmdWithOutput(t, "rbi", exportedFilePath)
+	var importResp string
+	if !assert.NoError(t, json.Unmarshal([]byte(output), &importResp)) {
+		return
+	}
+	assert.Equal(t, "", importResp)
```
RobiNino, line 462: Checking the output is empty is a bit pointless and doesn't indicate success. 
Rather focus on verifying the release bundle is actually in Artifactory. You can do it using `getReleaseBundleSpecification` for example (assert non existing before import, existing after)

utils/cliutils/commandsflags.go
```diff
@@ -2024,6 +2025,9 @@ var commandFlags = map[string][]string{
 		platformUrl, user, password, accessToken, serverId, lcPathMappingTarget, lcPathMappingPattern, Project,
 		downloadMinSplit, downloadSplitCount,
 	},
+	ReleaseBundleImport: {
+		user, password, accessToken, serverId, lcDryRun,
```
RobiNino, line 2029: Looks like we are not actually supporting the dry run capability and this flag isn't used. Lets add the functionality throughout the code.
Also, missing `platformUrl` here

RobiNino (changes requested):

lifecycle_test.go
EyalDelarea: Custom file name does work, the import process check the content of the file and it has to match, i will add a test for that to be on the safe side 🚀

artifactory_test.go
```diff
@@ -224,6 +224,21 @@ func TestArtifactorySimpleUploadSpecUsingConfig(t *testing.T) {
...
+	sendArtifactoryTrustedPublicKey(artHttpDetails)
+	// Import the release bundle
+	wd, err := os.Getwd()
+	assert.NoError(t, err)
+	testFilePath := filepath.Join(wd, "testdata", "lifecycle", "import", "cli-tests-2.zip")
+	assert.NoError(t, lcCli.Exec("rbi", testFilePath))
+
+	// Cleanup
+	deleteReceivedReleaseBundle("cli-tests", "2")
+	cleanArtifactoryTest()
```
RobiNino, line 240: Best to `defer cleanArtifactoryTest()` in the beginning of the test in case we do not reach the end of the function.

artifactory_test.go
```diff
@@ -5807,3 +5822,29 @@ func downloadModuleAndVerify() clientutils.ExecutionHandlerFunc {
...
+	resp, body, err := client.SendPost(*tests.JfrogUrl+"artifactory/api/security/keys/trusted", []byte(requestBody), artHttpDetails, "")
+	if err = errorutils.CheckResponseStatusWithBody(resp, body, http.StatusCreated, http.StatusConflict); err != nil {
+		log.Error(err.Error())
+		os.Exit(1)
+	}
+}
+
+func deleteReceivedReleaseBundle(bundleName, bundleVersion string) {
+	client, err := httpclient.ClientBuilder().Build()
+	coreutils.ExitOnErr(err)
```
RobiNino, line 5843: Since this is the end of the test, maybe we can just `assert.NoError()` here instead of exiting (which will prevent the clean up).

artifactory_test.go
```diff
@@ -5807,3 +5822,29 @@ func downloadModuleAndVerify() clientutils.ExecutionHandlerFunc {
...
+	deleteApi := path.Join("artifactory/api/release/bundles/", bundleName, bundleVersion)
+	resp, body, err := client.SendDelete(*tests.JfrogUrl+deleteApi, []byte{}, artHttpDetails, "Deleting release bundle")
+	if err = errorutils.CheckResponseStatusWithBody(resp, body, http.StatusOK); err != nil {
+		log.Error(err.Error())
```
RobiNino, line 5847: Same comment as above

artifactory_test.go
```diff
@@ -5807,3 +5822,29 @@ func downloadModuleAndVerify() clientutils.ExecutionHandlerFunc {
...
+		os.Exit(1)
+	}
+}
+
+func deleteReceivedReleaseBundle(bundleName, bundleVersion string) {
+	client, err := httpclient.ClientBuilder().Build()
+	coreutils.ExitOnErr(err)
+	deleteApi := path.Join("artifactory/api/release/bundles/", bundleName, bundleVersion)
+	resp, body, err := client.SendDelete(*tests.JfrogUrl+deleteApi, []byte{}, artHttpDetails, "Deleting release bundle")
```
RobiNino, line 5845: Error not checked

RobiNino (changes requested):

PR #2477, opened 2024-03-14
Deprecate the 'jf rt atc' command

yahavi:

---

The `/artifactory/api/security/token` endpoint is deprecated in Artifactory.
Deprecate `jf rt atc` and suggest using `jf atc` instead. `jf atc` is using the `/access/api/v1/tokens` API.

```
./jf rt atc
10:54:27 [🟠Warn] You are using a deprecated syntax of the command.
	Instead of:
	$ jf rt atc ...
	Use:
	$ jf atc ...
```

Downside - The `jf atc` does not support basic auth:
```
./jf atc
10:55:03 [🚨Error] authenticating with access token is currently mandatory for creating access tokens
```

### DIFF
--- a/artifactory/cli.go
+++ b/artifactory/cli.go
@@ -885,6 +885,7 @@ func GetCommands() []cli.Command {
 		},
 		{
 			Name:         "access-token-create",
+			Hidden:       true,
 			Aliases:      []string{"atc"},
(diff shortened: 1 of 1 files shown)


StephenWithPH:
@yahavi I'm seeing other bad behavior regarding auth (similar to your comment above) with `jf atc` now that it has replaced `jf rt atc`. For example, #2490 .

PR #2518, opened 2024-04-11
Support tracing all requests by a command

RobiNino:

---
Following this PR, every command will have a trace ID that will be logged and will allow users to easily identify which logs on the server side are related to the command executed by the CLI.

Depends on https://github.com/jfrog/jfrog-client-go/pull/937


main.go
```diff
@@ -136,12 +140,40 @@ func execMain() error {
...
+func generateTraceIdToken() (string, error) {
+	// Generate 8 random bytes.
+	var buf [8]byte
+	_, err := rand.Read(buf[:])
+	if err != nil {
+		return "", errorutils.CheckError(err)
+	}
+
+	// Convert the random bytes to a 16 chars hexadecimal string.
+	return hex.EncodeToString(buf[:]), nil
```
yahavi, line 174: The `[:]` creates a new slice from an array. Here is a solution without creating new slices:
```suggestion
	// Generate 8 random bytes.
	buf := make([]byte, 8)
	_, err := rand.Read(buf)
	if err != nil {
		return "", errorutils.CheckError(err)
	}

	// Convert the random bytes to a 16 chars hexadecimal string.
	return hex.EncodeToString(buf), nil
```

main.go
```diff
@@ -136,12 +140,40 @@ func execMain() error {
 		if warningMessage != "" {
 			clientlog.Warn(warningMessage)
 		}
+		if err = setUberTraceIdToken(); err != nil {
+			clientlog.Debug("failed generating a trace ID token:", err.Error())
```
yahavi, line 144: Do we expect it to fail sometime? If not, shouldn't it be a warning?

yahavi (changes requested):

Issue #2535, opened 2024-05-01
Release Bundle Create command (jf rbc) has an issue with the --project flag when using a spec file that is not "builds" but "pattern" or "aql"


derekscp:
### Describe the bug

When utilizing the 'jf release-bundle-create' [command](https://docs.jfrog-applications.jfrog.io/jfrog-applications/jfrog-cli/cli-for-jfrog-artifactory/release-lifecycle-management#creating-a-release-bundle) it results in an error when using 'aql' or 'pattern' in the spec file while there is a --project flag. Using 'builds' does not produce the error.
Error:
[🚨Error] release bundles creation source only supports the 'exclusions', 'props', 'excludeProps' and 'recursive' fields

### Current behavior

We get this error because of the project key. If it does not have the project key it works as expected.

$ jf release-bundle-create --server-id=cloud  --project=<PROJECT_KEY> --signing-key="gpgkey1" --sync=true --spec=specfile.json Release-Bundle-Test 1.0.0
16:38:03 [Debug] JFrog CLI version: 2.56.0
16:38:03 [Debug] OS/Arch: darwin/amd64
16:38:03 [Debug] Sending HTTP GET request to: <ART_URL>/artifactory/api/system/version
16:38:03 [Debug] Usage Report: Sending info...
16:38:03 [Debug] Sending HTTP GET request to: <ART_URL>/artifactory/api/system/version
16:38:03 [Debug] Artifactory response: 200 
16:38:03 [Debug] Artifactory response: 200 
16:38:03 [Debug] JFrog Artifactory version is: 7.84.3
16:38:03 [Debug] Sending HTTP POST request to: <ART_URL>/artifactory/api/system/usage
16:38:03 [🚨Error] release bundles creation source only supports the 'exclusions', 'props', 'excludeProps' and 'recursive' fields

### Reproduction steps

1. Create a specfile, specfile.json
```
{
  "files": [
    {
      "pattern": "<REPO_NAME>/<PATH_TO_FOLDER>/*"
    }
  ]
}
``` 
2. Run the jfrog cli command to create the release bundle under a project
```
$ jf release-bundle-create --server-id=cloud --project=<PROJECT_KEY> --signing-key="gpgkey1" --sync=true --spec=specfile.json Release-Bundle-Test 1.0.0
``` 

### Expected behavior

I would expect that it should work the same as when the spec file contains a build as the parameter instead of a pattern or aql.

### JFrog CLI version

jf version 2.56.0

[...]

PR #2536, opened 2024-05-01
Fix release bundle creation with project

RobiNino:

---
Following https://github.com/jfrog/jfrog-cli/issues/2535
Tests depend on https://github.com/jfrog/jfrog-client-go/pull/948


utils/cliutils/utils.go
```diff
@@ -401,19 +401,32 @@ func handleSecretInput(c *cli.Context, stringFlag, stdinFlag string) (secret str
...
 
-func GetSpec(c *cli.Context, isDownload bool) (specFiles *speccore.SpecFiles, err error) {
+func GetSpec(c *cli.Context, isDownload, overrideFieldsIfSet bool) (specFiles *speccore.SpecFiles, err error) {
 	specFiles, err = speccore.CreateSpecFromFile(c.String("spec"), coreutils.SpecVarsStringToMap(c.String("spec-vars")))
 	if err != nil {
 		return nil, err
 	}
-	// Override spec with CLI options
+	trimPatternPrefixIfDownload(specFiles, isDownload)
+	overrideSpecFieldsIfNeeded(c, specFiles, overrideFieldsIfSet)
```
yahavi, line 410: I find this a bit more simple to the reader:
```suggestion
	if isDownload {
		trimPatternPrefix(specFiles)
	}
	if overrideFieldsIfSet {
		overrideSpecFields(c, specFiles)
	}
```

lifecycle/cli.go
```diff
@@ -155,12 +155,9 @@ func create(c *cli.Context) (err error) {
 		return err
 	}
 
-	var creationSpec *spec.SpecFiles
-	if c.IsSet("spec") {
-		creationSpec, err = cliutils.GetSpec(c, true)
-		if err != nil {
-			return
-		}
+	creationSpec, err := getReleaseBundleCreationSpec(c)
```
yahavi, line 158: Instead of dealing with nil specs and dealing with flags in the core, can we create a default spec like we do in Artifactory commands?
For example - 
https://github.com/jfrog/jfrog-cli/blob/v2.56.0/artifactory/cli.go#L1188

yahavi (changes requested):

lifecycle/cli.go
RobiNino: There isn't really a default spec here because the specs depend on the creation method. In addition, the default spec is convenient when the user can provide options instead of a spec, which isn't currently the case.

PR #2573, opened 2024-06-11
Upgrade the Ubuntu version used to create JFrog CLI's debian package

eyalbe4:
Use the ubuntu:20.04 instead of ubuntu:16.04 to create JFrog CLI's debian package.


sverdlov93 (commented):
why 20.04 and not 24.04?

eyalbe4:
It may be best to upgrade Ubuntu gradually. It's hard to be absolutely sure whether  there will be some unintended implications on the debian packages that will be built on top of that platform @sverdlov93.

PR #2639, opened 2024-08-05
JFrog Curation - Support "Implicit curation"  - Run curation after "install" command failure

asafambar:
- [X] All [tests](https://github.com/jfrog/jfrog-cli/CONTRIBUTING.md#tests) have passed. If this feature is not already covered by the tests, new tests have been added.
- [X] The pull request is targeting the `dev` branch.
- [X] The code has been validated to compile successfully by running `go vet ./...`.
- [X] The code has been formatted properly using `go fmt ./...`.

---


buildtools/cli.go
```diff
@@ -391,6 +402,25 @@ func GetCommands() []cli.Command {
 	})
 }
 
+func wrapCmdWithCurationPostFailureRun(c *cli.Context, cmd func(c *cli.Context) error) error {
```
eyalbe4, line 405: This function should be placed in the jfrog-cli-security module.

buildtools/cli.go
```diff
@@ -391,6 +402,25 @@ func GetCommands() []cli.Command {
...
+	if err := cmd(c); err != nil {
+		CurationInspectAfterFailure(c, err)
+		return err
+	}
+	return nil
+}
+
+func CurationInspectAfterFailure(c *cli.Context, errFromCmd error) {
```
eyalbe4, line 413: This function should be placed in the jfrog-cli-security module.

eyalbe4 (changes requested):

buildtools/cli.go
```diff
@@ -325,9 +335,13 @@ func GetCommands() []cli.Command {
 			SkipFlagParsing: true,
 			BashComplete:    corecommon.CreateBashCompletionFunc("install", "i", "isntall", "add", "ci", "publish", "p"),
 			Category:        buildToolsCategory,
-			Action: func(c *cli.Context) error {
+			Action: func(c *cli.Context) (errFromCmd error) {
 				cmdName, _ := getCommandName(c.Args())
-				return npmGenericCmd(c, cmdName, false)
+				if errFromCmd = npmGenericCmd(c, cmdName, false); errFromCmd != nil {
```
eyalbe4, line 340: This if statement that checks whether the error is nil or not should be placed inside the jfrog-cli-security module. This is to reduce the dependecy of jfrog-cli-security on this module, in case changes are needed to the curation logic.
The approach taken for "pip" is better, so let's adopt it here as well - 
image

eyalbe4 (changes requested):
Does jfrog-cli-security include a mechanism to check whether the user has JFrog Curation enabled, and if so skip the execution for the "jf curation-audit" command?

buildtools/cli.go
asafambar: its a bit more complicated, I will check again

PR #2686, opened 2024-09-05
Fail 'jf docker scan/push/pull' when a wrong flag is provided to the command

eranturgeman:

---
When a wrong flag is provided to the command it is captured as an argument and therefore is not going through flags validations. This causes the flag to be ignored and without indications on anything wrong.
Since 'jf docker scan/push/pull' expects 2 arguments only: 'docker' + one of the commands, and extra argument indicates wrong usage.
I added a validation on the arguments number to capture this issue

### DIFF
--- a/buildtools/cli.go
+++ b/buildtools/cli.go
@@ -3,6 +3,7 @@ package buildtools
 import (
 	"errors"
 	"fmt"
+	"github.com/jfrog/jfrog-cli-security/utils"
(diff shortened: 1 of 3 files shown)


eyalbe4 (changes requested):
@eranturgeman,
Will you be able fix the jfrog-cli-security version in the go.mod, to allow the tests to pass?

PR #2708, opened 2024-09-30
Cocoapods config

barv-jfrog:

---

### DIFF
--- a/utils/cliutils/commandsflags.go
+++ b/utils/cliutils/commandsflags.go
@@ -37,6 +37,7 @@ const (
 	GitLfsClean            = "git-lfs-clean"
 	Mvn                    = "mvn"
 	MvnConfig              = "mvn-config"
+	CocoapodsConfig        = "cocoapods-config"
 	Gradle                 = "gradle"
 	GradleConfig           = "gradle-config"
 	DockerPromote          = "docker-promote"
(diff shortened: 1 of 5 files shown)


utils/cliutils/commandsflags.go
```diff
@@ -1815,6 +1816,9 @@ var commandFlags = map[string][]string{
 		url, user, password, accessToken, sshPassphrase, sshKeyPath, serverId, refs, glcRepo, glcDryRun,
 		glcQuiet, InsecureTls, retries, retryWaitTime,
 	},
+	CocoapodsConfig: {
+		global, serverIdResolve, serverIdDeploy, repoResolveReleases, repoResolveSnapshots, repoDeploySnapshots, repoDeployReleases,
```
eyalbe4, line 1820: `serverIdDeploy` shouldn't be added for package managers that do not deploy packages to Artifactory.
In addition, the following options should be removed (they are all maven specific):
* repoResolveReleases
* repoResolveSnapshots
* repoDeploySnapshots

The only options that are needed are the below (see `NugetConfig` above as a reference):
* global
* serverIdResolve
* repoResolve

eyalbe4 (changes requested):

utils/cliutils/commandsflags.go
srmish-jfrog: Fixed

PR #2730, opened 2024-10-20
Fix integration tests

yahavi:

---

Update the tests to align with changes introduced in Artifactory 7.90+:

* REST APIs to `http://127.0.0.1:8081/access` are no longer functional. Instead, use `http://127.0.0.1:8082/access`.
* The archive indexing feature has been [completely removed](https://jfrog.com/help/r/artifactory-archive-search-depreciation/artifactory-archive-search-depreciation), meaning we cannot test this feature in the newer versions of Artifactory. In this pull request, the archive indexing tests have been entirely removed.
* Make the Access tests to run locally.

Additionally, in this pull request, the access tests are now executed locally on 127.0.0.1 instead of the SaaS server.


eyalbe4 (changes requested):
I assume that the removal of the archive indexing also means that the below flag no longer works from a specific version of Artifactory.
If this is the case, let's also add the following prefix to the flag's description -

*This option stopped being supported since version XXX for Artifactory.* 
(replace XXX with the right version). 
Let's also update the online documentation with that description.

```
archiveEntries: cli.StringFlag{
		Name:  archiveEntries,
		Usage: "[Optional] If specified, only archive artifacts containing entries matching this pattern are matched. You can use wildcards to specify multiple artifacts.` `",
	},
```

PR #2742, opened 2024-10-31
Fix setup of GitLab Templates for JFrog for air-gapped environments.

florianbehrens:
Using GitLab Templates for JFrog did not work as described [here](https://docs.jfrog-applications.jfrog.io/ci-and-sdks/ci-integrations/gitlab-templates-for-jfrog) for "air-gapped" environments. The reason seems to be a bug in build/gitlab/v2/.setup-jfrog-unix.yml.

Only build/gitlab/v2/.setup-jfrog-unix.yml was changed and I tested it on my local GitLab/Artifactory instances.


florianbehrens:

PR #2762, opened 2024-11-17
Make the signing key optinal `jf release-bundle-create`

oshratZairi:

---
Turn the signing key flag to optional for RBC and in RBP command. 
Documentation PR https://github.com/jfrog/documentation/pull/172


lifecycle_test.go
```diff
@@ -363,11 +377,12 @@ func uploadBuildWithDeps(t *testing.T, buildName, buildNumber string) {
 	runRt(t, "build-publish", buildName, buildNumber)
 }
 
-func initLifecycleTest(t *testing.T) (cleanCallback func()) {
+func initLifecycleTest(t *testing.T, minVersion string) (cleanCallback func()) {
```
eyalbe4, line 380: There's probably no need to pass `minVersion` as an argument for this function, because the `signingKeyOptionalArtifactoryMinVersion` const should already be accessible for it.

lifecycle_test.go
```diff
@@ -195,22 +206,25 @@ func uploadBuilds(t *testing.T) func() {
...
-func createRbFromSpec(t *testing.T, specName, rbName, rbVersion string, sync bool) {
+func createRbFromSpec(t *testing.T, specName, rbName, rbVersion string, sync bool, withoutKey bool) {
 	specFile, err := tests.CreateSpec(specName)
 	assert.NoError(t, err)
-	createRb(t, specFile, "spec", rbName, rbVersion, sync)
+	createRb(t, specFile, "spec", rbName, rbVersion, sync, withoutKey)
 }
 
-func createRb(t *testing.T, specFilePath, sourceOption, rbName, rbVersion string, sync bool) {
+func createRb(t *testing.T, specFilePath, sourceOption, rbName, rbVersion string, sync bool, withoutKey bool) {
```
eyalbe4, line 218: withoutKey --> withoutSigningKey

(to make it easier for the random reader to understand what this argument is).

utils/cliutils/commandsflags.go
```diff
@@ -1650,7 +1650,7 @@ var flagsMap = map[string]cli.Flag{
 	},
 	lcSigningKey: cli.StringFlag{
 		Name:  SigningKey,
-		Usage: "[Mandatory] The GPG/RSA key-pair name given in Artifactory.` `",
+		Usage: "[Optional] The GPG/RSA key-pair name given in Artifactory. If not provided will create/use the default one` `",
```
eyalbe4, line 1653: [Optional] The GPG/RSA key-pair name given in Artifactory. If not provided will create/use the default one
-->
[Optional] The GPG/RSA key-pair name given in Artifactory. If the key isn't provided, the command creates or uses the default key

Let's also have a PR with this change opened to [github.com/jfrog/documentation](https://github.com/jfrog/documentation) and link the PR to this PR by adding to the PR description. [github.com/jfrog/documentation](https://github.com/jfrog/documentation) includes the source files for JFrog CLI's online documentation.
We will merge this PR documentation once this functionality is released.

eyalbe4 (changes requested):

lifecycle_test.go
oshratZairi: The initLifecycleTest function is used across multiple tests. However, for the signingKey test, I need to ensure it includes the minimum RT version required to support this API, as this parameter is necessary.

eyalbe4: @oshratZairi,
The `initLifecycleTest` function is used only from within the `lifecycle.go` file.
This function can access the `signingKeyOptionalArtifactoryMinVersion` const on its own, and therefore there's no need for the function to recieve it as an argument.

PR #2772, opened 2024-11-27
Enhancements to Release Bundle Create Command: Optional Flags and Fallback Support

oshratZairi:

---
Improvements to the release bundle create command:
- Made the `spec` flag optional, with `build-name` and `build-number` flags as replacements.
- Added support for `JFROG_CLI_BUILD_NAME` and `JFROG_CLI_BUILD_NUMBER` as fallback options when the `build-name` and `build-number` flags are not provided.

documentation - https://github.com/jfrog/documentation/pull/173/files


utils/cliutils/commandsflags.go
```diff
@@ -579,6 +579,8 @@ const (
 	ReleaseBundles       = "release-bundles"
 	lcReleaseBundles     = lifecyclePrefix + ReleaseBundles
 	SigningKey           = "signing-key"
+	BuildName            = "build-name"
+	BuildNumber          = "build-number"
```
eyalbe4, line 583: No need to define the above two consts.
You should reuse the existing `buildName` and `buildNumber` consts defined above.

utils/cliutils/utils.go
```diff
@@ -415,6 +415,10 @@ func GetSpec(c *cli.Context, isDownload, overrideFieldsIfSet bool) (specFiles *s
 	return
 }
 
+func CreateSpecFromBuildNameAndNumber(buildName string, buildNumber string) (specFiles *speccore.SpecFiles, err error) {
```
eyalbe4, line 418: This wrapper function seema redundant. The `speccore.CreateSpecFromBuildNameAndNumber` can be called directly.

utils/tests/utils.go
```diff
@@ -693,3 +693,15 @@ func createFlagSet(t *testing.T, flags []string, args []string) *flag.FlagSet {
 	assert.NoError(t, flagSet.Parse(cmdFlags))
 	return flagSet
 }
+
+func ExtractBuildInfo(build string) (string, string, error) {
```
eyalbe4, line 697: You should not recreate this split logic.
JFrog CLI already splits the build value when it actually uses the spec file to download files. You should find that utility and reuse it. Note that code may be in jfrog-client-go or jfrog-cli-core.

lifecycle_test.go
```diff
@@ -167,7 +169,24 @@ func TestLifecycleFullFlow(t *testing.T) {
 		// Verify the artifacts were distributed correctly by the provided path mappings.
 		assertExpectedArtifacts(t, tests.SearchAllDevRepo, tests.GetExpectedLifecycleDistributedArtifacts())
 	*/
+}
+
+func TestCreateBundleWithoutSpec(t *testing.T) {
+	cleanCallback := initLifecycleTest(t, artifactoryLifecycleMinVersion)
+	defer cleanCallback()
+
+	lcManager := getLcServiceManager(t)
```
eyalbe4, line 178: Better to move this function call a few lines below, closer to where it is needed.

eyalbe4 (changes requested):

utils/tests/utils.go
oshratZairi: the logic here is just for the test - I need to add the --build-name, --build-number to the command according to the test files and then will run the cli command. The create command will run as expected with no external locig.

utils/cliutils/commandsflags.go
```diff
@@ -164,8 +164,8 @@ const (
 	specVars = "spec-vars"
 
 	// Build info flags
-	buildName   = "build-name"
-	buildNumber = "build-number"
+	BuildName   = "build-name"
+	BuildNumber = "build-number"
```
eyalbe4, line 168: Why did we make these variables public? They are used only in this file?

eyalbe4 (changes requested):

lifecycle/cli.go
```diff
@@ -131,21 +134,56 @@ func validateCreateReleaseBundleContext(c *cli.Context) error {
...
+		return err
+	}
+
+	if err := validateCreationValuesPresence(c, methodCount); err != nil {
+		return err
+	}
+	return nil
+}
+
+func validateSingleCreationMethod(methodCount int) error {
```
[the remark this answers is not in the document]
oshratZairi: this function verifies that the user didn't send  2 sources for the bundle creation for example --spec and --builds
the methodCount isn't redundant - the value for the arguments comes from the check in line 146.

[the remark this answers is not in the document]
eyalbe4: :+1:

utils/cliutils/commandsflags.go
oshratZairi: You have asked me to use BuildName and BuildNumber and not to add new ones to the command - so had to turn the current into public

eyalbe4: Got it. I didn't notice those flags are used from within a different package,

PR #2805, opened 2024-12-26
Enhance download functionality to support RBV2

oshratZairi:

---
Enhance download functionality to support RBV2


go.mod
```diff
@@ -168,6 +168,10 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
+replace github.com/jfrog/jfrog-cli-core/v2 => ../jfrog-cli-core
```
EyalDelarea, line 171: Notice to change replaces before pushing :)

EyalDelarea (changes requested):

artifactory/cli.go
```diff
@@ -1286,10 +1295,23 @@ func prepareDownloadCommand(c *cli.Context) (*spec.SpecFiles, error) {
 }
 
 func downloadCmd(c *cli.Context) error {
-	downloadSpec, err := prepareDownloadCommand(c)
+	var isRbv2 bool
+	var err error
+	if c.IsSet("bundle") {
+		isRbv2, err = checkRbExistenceInV2(c)
+	}
+	//if the command contains --bundle and the bundle found in rbv2, we will download from rbv2
```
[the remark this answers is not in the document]
oshratZairi: the logic is:
if we have the same rb-name in both rbv1 and rbv2 - we will take rbv2.
if the cmd dose not contain --bundle ,  it will remain as it is. 
moved the logic.

artifactory/cli.go
```diff
@@ -2648,8 +2702,27 @@ func createDefaultDownloadSpec(c *cli.Context) (*spec.SpecFiles, error) {
...
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		source = strings.TrimPrefix(c.Args().Get(0), "/")
+	}
+
 	return spec.NewBuilder().
-		Pattern(strings.TrimPrefix(c.Args().Get(0), "/")).
+		Pattern(source).
```
EyalDelarea, line 2725: Can we put all the source logic inside a function that will also explain why it should be different?

Something like this:
```go
func getSourcePattern(c *cli.Context) (string, error) {
	if c.IsSet("bundle") {
		isRbv2, err := checkRbExistenceInV2(c)
		if err != nil {
			log.Debug("Error occurred while checking if the bundle exists in rbv2:", err.Error())
		} else if isRbv2 {
			return buildSourceForRbv2(c)
		}
	}
	return strings.TrimPrefix(c.Args().Get(0), "/"), nil
}
```

then the line will look like this:

```go
return spec.NewBuilder().
		Pattern(getSourcePattern(c)).

**Review by EyalDelarea — CHANGES_REQUESTED** [review]

**Comment by EyalDelarea** [comment]
@oshratZairi looks good!

Just to extract the last function and fix the static check and we are good to go 👍

File: artifactory/cli.go [hunk 4]
```diff
@@ -2644,12 +2698,23 @@ func createGitLfsCleanConfiguration(c *cli.Context) (gitLfsCleanConfiguration *g
 }
 
 func createDefaultDownloadSpec(c *cli.Context) (*spec.SpecFiles, error) {
+	var isRbv2 bool
+
 	offset, limit, err := getOffsetAndLimitValues(c)
 	if err != nil {
 		return nil, err
 	}
+
```
EyalDelarea, line 2707: this could also move inside the `getSourcePattern` function

EyalDelarea (approved):

PR #2866, opened 2025-02-10
Validate `Yarn` Version and Improve Error Messaging for Unsupported Versions

fluxxBot:

---

depends on:-
1. https://github.com/jfrog/jfrog-cli-core/pull/1347
2. https://github.com/jfrog/jfrog-cli-artifactory/pull/44


npm_test.go
```diff
@@ -714,6 +714,45 @@ func TestYarn(t *testing.T) {
 	inttestutils.DeleteBuild(serverDetails.ArtifactoryUrl, tests.YarnBuildName, artHttpDetails)
 }
 
+func TestYarnUpgradeToV4(t *testing.T) {
```
EyalDelarea, line 717: Let's also add a test the verify setting a correct version works.

EyalDelarea (changes requested):
Please update the test cases to address new changes made to the other PRs.

Also i'm editing the PR title, as CLI pull requests will appear in the release notes of the CLI, and this is the place to notify users about the improvement that has been introduced here. ( also by the core and artifactory )

.github/workflows/npmTests.yml
```diff
@@ -41,4 +41,6 @@ jobs:
           RTLIC: ${{ secrets.RTLIC }}
 
       - name: Run npm tests
+        env:
```
fluxxBot, line 44: this is required since yarn4 is not compatible with node16, this env will allow yarn4 tests to run in the same env.
Note: upgrading node16 to node18 causes legacyNpm tests to fail

EyalDelarea: Do we need all the yarn v4 tests now that we removed the support for yarnv4 commands?

PR #2873, opened 2025-02-13
Introduced overwrite flag for build publish command which overwrites builds with same number

bhanurp:

---
- An optional --overwrite flag is added to build publish command to overwrite build info with same build number.
- Artifacts wont be affected


buildinfo_test.go
```diff
@@ -658,6 +658,73 @@ func TestBuildAddGitEnvBuildNameAndNumber(t *testing.T) {
...
+	// Publish build info with overwrite flag
+	runRt(t, "bp", buildName, buildNumber, "--overwrite=true")
+	publishedBuildInfo, found, err = tests.GetBuildRuns(serverDetails, buildName, buildNumber)
+	assert.NoError(t, err)
+	assert.True(t, found, "build info was expected to be found")
+	buildNumbersFrequencyAfterOverwrite := calculateBuildNumberFrequency(publishedBuildInfo)
+	// Since overwrite is ran for buildNumber verify if only one build info is available for given build number
+	assert.Equal(t, 1, buildNumbersFrequencyAfterOverwrite[buildNumber])
+	// Since overwrite is ran for buildNumber verify no change for preReleaseBuildNumber
+	assert.Equal(t, 2, buildNumbersFrequencyAfterOverwrite[preReleaseBuildNumber])
```
RobiNino, line 698: This code block is kind of duplicated for every test case, maybe a common function would be useful

buildinfo_test.go
```diff
@@ -658,6 +658,73 @@ func TestBuildAddGitEnvBuildNameAndNumber(t *testing.T) {
 	testBuildAddGit(t, true)
 }
 
+func TestBuildPublishWithOverwrite(t *testing.T) {
+	initArtifactoryTest(t, "")
+	buildName := "overwrite-test-build"
```
RobiNino, line 663: I suggest using `tests.RtBuildName1` as the build number. The test suite initializes it with a unique suffix (OS & timestamp) to avoid collisions with parallel test runs.

buildinfo_test.go
```diff
@@ -737,6 +804,15 @@ func testBuildAddGit(t *testing.T, useEnvBuildNameAndNumber bool) {
 	cleanArtifactoryTest()
 }
 
+func calculateBuildNumberFrequency(runs *buildinfo.BuildRuns) map[string]int {
```
RobiNino, line 807: Perhaps we can use the utility from `jfrog-cli-artifactory` to avoid duplication

utils/tests/utils.go
```diff
@@ -240,6 +240,17 @@ func GetBuildInfo(serverDetails *config.ServerDetails, buildName, buildNumber st
 	return servicesManager.GetBuildInfo(params)
 }
 
+func GetBuildRuns(serverDetails *config.ServerDetails, buildName, buildNumber string) (pbi *buildinfo.BuildRuns, found bool, err error) {
```
RobiNino, line 243: This API ignore the buildNumber, see my comment on the `jfrog-cli-artifactory` PR.

utils/cliutils/commandsflags.go
```diff
@@ -1782,7 +1782,7 @@ var commandFlags = map[string][]string{
 	},
 	BuildPublish: {
 		url, user, password, accessToken, sshPassphrase, sshKeyPath, serverId, buildUrl, bpDryRun,
-		envInclude, envExclude, InsecureTls, Project, bpDetailedSummary,
+		envInclude, envExclude, InsecureTls, Project, bpDetailedSummary, Overwrite,
```
RobiNino, line 1785: Since we are using the existing `Overwrite` flag, the current description will be shown in the help:
`[Default: false] Overwrites the instance configuration if an instance with the same ID already exists.`

I suggest introducing a new flag in the `flagsMap` with an accurate description of the functionality it adds. It can have a different index in the map, but the same flag name.

RobiNino (changes requested):

buildinfo_test.go
bhanurp: I tried moving lot of them into a common function and when I read test case whats actually happening is abstracted into common function making it less readable how ever I moved the common part of assert.NoError and assert.True into a common function which is very much readable

utils/cliutils/commandsflags.go
```diff
@@ -743,6 +744,10 @@ var flagsMap = map[string]cli.Flag{
 		Name:  Overwrite,
 		Usage: "[Default: false] Overwrites the instance configuration if an instance with the same ID already exists.` `",
 	},
+	bpOverwrite: cli.BoolFlag{
+		Name:  Overwrite,
+		Usage: "[Default: false] Overwrites existing build info with build name and number.` `",
```
RobiNino, line 749: I suggest adding a bit more info about the actual behavior, for example:
```suggestion
		Usage: "[Default: false] Overwrites all existing occurrences of build infos with the provided name and number. Build artifacts will not be deleted.` `",
```

RobiNino (approved):

RobiNino:
Perhaps we should also add a simple dry run test, or enhance the existing `TestBuildPublishDryRun`

bhanurp:
> Perhaps we should also add a simple dry run test, or enhance the existing `TestBuildPublishDryRun`

Updated TestBuildPublishDryRun with test for overwrite

PR #2926, opened 2025-03-17
Allow Build Publish with `overwrite` Flag When Build Number is New

bhanurp:

---
## Depends On
https://github.com/jfrog/jfrog-cli-artifactory/pull/55

---
### **Issue**
When attempting to publish a build with the `overwrite` flag, the process fails if the `buildName` already exists in Artifactory but the `buildNumber` is new. The failure occurs due to a missing build number in Artifactory.

### **Current Behavior**
- Publishing fails with an error stating that the build number is not available.

### **Solution**
- Ensure that the build info is pushed even if the specified `buildNumber` is not yet available in Artifactory.

### DIFF
--- a/buildinfo_test.go
+++ b/buildinfo_test.go
@@ -4,11 +4,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/formats"
 	clientTestUtils "github.com/jfrog/jfrog-client-go/utils/tests"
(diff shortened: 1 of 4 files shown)


buildinfo_test.go
```diff
@@ -728,6 +729,12 @@ func TestBuildPublishWithOverwrite(t *testing.T) {
 	// Verify even though overwrite is used when no build infos are available build info should be published
 	assertBuildNumberOccurrencesForGivenBuildNameAndNumber(t, publishedBuildInfo, 1, found, buildNumber, err)
 
+	// Run build-publish with overwrite flag and build should be published
+	runRt(t, "bp", buildName, nonExistingBuildNumber, "--overwrite=true")
+	publishedBuildInfo, found, err = tests.GetBuildRuns(serverDetails, buildName)
```
itsmeleela, line 734: Do we have a test already that is testing same scenario but flag value is false (I mean without flag)?

I suggest to add one as we are here anyway

bhanurp: by default all other tests doesn't include flag(--overwrite) which makes default (--overwrite=false)

PR #2971, opened 2025-05-11
Feature/docker validate sha

agrasth:

---
# Add `--validate-sha` flag for Docker push command

## Description
This PR adds a new `--validate-sha` flag to the Docker push command. When enabled, the flag allows the CLI to use the image's SHA digest for validation instead of the tag name during Docker push operations. This is particularly useful when pushing to virtual repositories where the tag might exist with different content in higher priority repositories.

## Changes
- Removed redundant direct flag registration in the Docker command
- Updated documentation with comprehensive explanation of the flag's purpose and behavior
- Added flag extraction in the push command

## Testing
- Manually tested the flag with both old (`jf rt docker-push`) and new (`jf docker push`) CLI syntax
- Verified behavior when pushing to repositories with existing tags
- Confirmed build info collection works correctly with the flag

## Related PRs
- jfrog-cli-artifactory: [here](https://github.com/jfrog/jfrog-cli-artifactory/pull/73)
- jfrog-cli-core: [here](https://github.com/jfrog/jfrog-cli-core/pull/1386)


docs/buildtools/dockerpush/help.go
```diff
@@ -8,5 +8,16 @@ func GetDescription() string {
 
 func GetArguments() string {
 	return `	docker push args
-		The docker push args to run docker push.`
+		The docker push args to run docker push.
+		
+	--validate-sha
```
bhanurp, line 13: since it is a flag, do we need to update here?

utils/cliutils/commandsflags.go
```diff
@@ -1708,6 +1710,11 @@ var flagsMap = map[string]cli.Flag{
 	runNative: cli.BoolFlag{
 		Name:  runNative,
 		Usage: "[Default: false] Set to true if you'd like to use the native client configurations. Note: This flag would invoke native client behind the scenes, has performance implications and does not support deployment view and detailed summary` `",
+	
+	},
+	validateSha: cli.BoolFlag{
```
bhanurp, line 1715: what ist he difference between container push command and docker, dockerPush is it like OCI support?

utils/cliutils/commandsflags.go
```diff
@@ -1845,7 +1852,7 @@ var commandFlags = map[string][]string{
 	},
 	ContainerPush: {
 		BuildName, BuildNumber, module, url, user, password, accessToken, sshPassphrase, sshKeyPath,
-		serverId, skipLogin, threads, Project, detailedSummary,
+		serverId, skipLogin, threads, Project, detailedSummary, validateSha,
```
bhanurp, line 1855: is the flag required here as well as in jfrog-cli-artifactory?

bhanurp (changes requested):

utils/cliutils/commandsflags.go
agrasth: What I understood was 
- DockerPush   Specifically for Docker images, using the Docker CLI under the hood
-  ContainerPush  - A more generic command that supports multiple container technologies, not just Docker.

Since both commands perform pushing operations that could benefit from SHA validation, it makes sense to add the flag to both.

agrasth: What I understood was 
- DockerPush   Specifically for Docker images, using the Docker CLI under the hood
-  ContainerPush  - A more generic command that supports multiple container technologies, not just Docker.

Since both commands perform pushing operations that could benefit from SHA validation, it makes sense to add the flag to both.

docs/buildtools/dockerpush/help.go
agrasth: Yeah, you are right, for flags we don't need to update here.

PR #2980, opened 2025-05-20
Add MCP server command

EyalDelarea:

---


general/mcp/cli.go
```diff
@@ -0,0 +1,443 @@
...
+	req, err := http.NewRequest(http.MethodHead, parsedURL.String(), nil)
+	if err != nil {
+		return "", err
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
```
asafgabai, line 365: Let's handle the returned error:
-->
```golang
	defer func() {
		if resp != nil && resp.Body != nil {
			err = errors.Join(err, errorutils.CheckError(resp.Body.Close()))
		}
	}()
```

general/mcp/cli.go
```diff
@@ -0,0 +1,443 @@
...
+
+	if !os.IsNotExist(err) {
+		return "", false, fmt.Errorf("failed to stat '%s': %w", fullPath, err)
+	}
+
+	return fullPath, false, nil
+}
+
+// downloadBinary downloads the binary from the remote server
+func downloadBinary(fullPath, version, osName, arch string) (string, error) {
```
asafgabai, line 272: Let's rename `fullPath` to something clearer like `targetPath` or `destPath`

general/mcp/cli.go
```diff
@@ -0,0 +1,443 @@
...
+	if err != nil {
+		return "", fmt.Errorf("invalid URL: %w", err)
+	}
+
+	// Ensure URL is using HTTPS
+	if parsedURL.Scheme != "https" {
+		return "", fmt.Errorf("URL must use HTTPS scheme")
+	}
+
+	resp, err := http.Get(parsedURL.String())
```
asafgabai, line 288: I believe we already have functions for downloading files/executables. Is there any such a function that we can use here?

general/mcp/cli.go
```diff
@@ -0,0 +1,443 @@
...
+	// Download the latest version
+	_, err = downloadBinary(fullPath, defaultServerVersion, osName, arch)
+	if err != nil {
+		return err
+	}
+
+	// Check new version after update
+	newVersion, err := getMcpServerVersion(fullPath)
+	if err != nil {
+		log.Warn("Could not determine new MCP server version:", err)
```
asafgabai, line 189: Wouldn't we want to return an error in this case?

general/mcp/cli.go
```diff
@@ -0,0 +1,443 @@
...
+	// Execute the command
+	return cmd.Run()
+}
+
+// createMcpServerCommand creates the exec.Command for the MCP server
+func createMcpServerCommand(executablePath, toolSets, toolAccess string) *exec.Cmd {
+	cmd := exec.Command(
+		executablePath,
+		cliutils.McpToolsets+toolSets,
+		cliutils.McpToolAccess+toolAccess,
```
asafgabai, line 103: Do these arguments really work? It looks like they are parsed to something like this:
```
./cli-mcp-server toolsetsall-toolsets tools-accessread
```
It doesn't seem right. Are these flags or arguments of the MCP server?

general/mcp/cli.go
```diff
@@ -0,0 +1,443 @@
...
+	cmd.Stderr = os.Stderr
+	cmd.Env = os.Environ()
+	return cmd
+}
+
+// logStartupInfo logs the MCP server startup parameters
+func logStartupInfo(toolSets, toolAccess string) {
+	displayToolset := toolSets
+	if displayToolset == "" {
+		displayToolset = defaultToolsets
```
asafgabai, line 116: What's the meaning of `defaultToolsets` and `defaultToolAccess` if they're only used locally in this function?

asafgabai (changes requested):

general/mcp/cli.go
EyalDelarea: We have tools to use specs file but this is from releases, we usually use it via curl.
In here, the requirements comes from the CLI itself.

general/mcp/cli.go
EyalDelarea: they are args which gets passed to the MCP server and they are important for visibility

EyalDelarea (commented):
I've removed the update command as i agree with you that when we will want to implement it we should do it right.

general/mcp/cli.go
EyalDelarea: They are args.

PR #2991, opened 2025-06-04
Added-ruby-config

basel1322:
Added ruby-config


buildtools/cli.go
```diff
@@ -127,6 +129,19 @@ func GetCommands() []cli.Command {
 				return cliutils.CreateConfigCmd(c, project.Gradle)
 			},
 		},
+		{
+			Name:         "ruby-config",
+			Aliases:      []string{"rubyc"},
+			Flags:        cliutils.GetCommandFlags(cliutils.RubyConfig),
+			Usage:        gradleconfig.GetDescription(),
```
RemiBou: why gradeconfig description ?

utils/cliutils/commandsflags.go
```diff
@@ -1821,6 +1824,9 @@ var commandFlags = map[string][]string{
 		global, serverIdResolve, serverIdDeploy, repoResolve, repoDeploy, usesPlugin, UseWrapper, deployMavenDesc,
 		deployIvyDesc, ivyDescPattern, ivyArtifactsPattern,
 	},
+	RubyConfig: {
```
RemiBou, line 1826: missing Ruby

buildtools/cli.go
basel1322: Fixed the typo

utils/cliutils/commandsflags.go
basel1322: We only want to add the ruby-config, should I remove the ruby  and only keep ruby-config ?

RemiBou: yes

PR #3165, opened 2025-10-15
Introduced support for docker build command

fluxxBot:

---

This PR aims to introduce support for docker build and buildx command while maintaing the legacy usage.

Implements docker build, docker buildx and legacy build commands using the Strategy pattern to support different execution modes based on the JFROG_RUN_NATIVE environment variable.

changes:
1. New Strategy Pattern Implementation: Three strategies for Docker build execution:
   -> RunNativeStrategy: Direct Docker build and buildx execution (JFROG_RUN_NATIVE=true)
   -> LegacyStrategy: Traditional JFrog approach (default)

depends on:
1. https://github.com/jfrog/jfrog-cli-artifactory/pull/283
2. https://github.com/jfrog/jfrog-client-go/pull/1269

### DIFF
--- a/.github/workflows/scriptTests.yml
+++ b/.github/workflows/scriptTests.yml
@@ -37,7 +37,7 @@ jobs:
             osSuffix: ""
 
           - name: macos
-            version: 13
+            version: 15
             osSuffix: ""
 
           - name: windows


buildtools/cli.go
```diff
@@ -848,6 +848,8 @@ func dockerCmd(c *cli.Context) error {
 		err = loginCmd(c)
 	case "scan":
 		return dockerScanCmd(c, image)
+	case "build", "buildx": // Handle both build and buildx with same handler
```
bhanurp:
```suggestion
     // Handle both build and buildx with same handler
	case "build", "buildx":
```

buildtools/cli.go
```diff
@@ -1041,6 +1074,43 @@ func extractDockerLoginOptionsFromArgs(args []string) (user, password string, er
...
+		dockerfilePath = "Dockerfile"
+	}
+
+	// Extract image tag from command
+	_, _, imageTag, err = coreutils.FindFlag("-t", args)
+	if err != nil || imageTag == "" {
+		// Try --tag flag as alternative
+		_, _, imageTag, _ = coreutils.FindFlag("--tag", args)
+	}
+	if imageTag == "" {
```
nitinp19, line 1145: is imageTag mandatory field why are we logging the error for imageTag only and not for other fields. If it is not a mandatory field we can skip the logging and if we really want to log it, we can actually log the actual error.

_, _, imageTag, err = coreutils.FindFlag("-t", args)
	if err != nil || imageTag == "" {
		// Try --tag flag as alternative
		_, _, imageTag, err = coreutils.FindFlag("--tag", args) 
        // log the error here 
	}

fluxxBot: other fields are not mandatory, this tag is

PR #3202, opened 2025-11-06
JGC-429 - Bump go version

ehl-jf:

---

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/jfrog/jfrog-cli
 
-go 1.24.6
+go 1.25.4
 
 replace (
 	// Should not be updated to 0.2.6 due to a bug (https://github.com/jfrog/jfrog-cli-core/pull/372)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ node("docker-ubuntu20-xlarge") {
     repo = 'jfrog-cli'
     sh 'rm -rf temp'
     sh 'mkdir temp'
-    def goRoot = tool 'go-1.24.6'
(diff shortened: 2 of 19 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/jfrog/jfrog-cli
 
-go 1.24.6
```
cyrilc-pro, line 3: This is the language level of the module, I think you can leave it as is.

ehl-jf: We need to bump to have security updates, right ?

cyrilc-pro:
I see you updated the GHA from https://github.com/jfrog/.github but when I discussed about this issue with @RemiBou, we said we should probably remove these GHA and let each project define its configuration for Go, Linters, etc.

ehl-jf:
> I see you updated the GHA from https://github.com/jfrog/.github but when I discussed about this issue with @RemiBou, we said we should probably remove these GHA and let each project define its configuration for Go, Linters, etc.

For now I've just follow the current bump go documentation.

I think is not a bad idea that every one use the same go version since are tightly interconnected for now

RemiBou (approved):

RemiBou:
> > I see you updated the GHA from https://github.com/jfrog/.github but when I discussed about this issue with @RemiBou, we said we should probably remove these GHA and let each project define its configuration for Go, Linters, etc.
> 
> For now I've just follow the current bump go documentation.
> 
> I think is not a bad idea that every one use the same go version since are tightly interconnected for now

I think we should remove this version from here, or make it configurable when using the step : the tooling version should sit next to the code in the repo, or else we can have issue like non reproductible builds for patch or conflicts. We'll do it in another piece of work

PR #3249, opened 2025-12-02
Fix for 'include' flag with spec file

agrasth:

---
# Fix: --include Flag Ignored When Using Spec Files in Search Command

## Description
Fixes bug where the `--include` flag was ignored when using spec files with `jf rt search`, causing all fields to be returned instead of only the specified fields.

### Before Fix:
```bash
jf rt search --spec=spec.json --include="name" --limit=1
# Returns: ALL fields (path, type, size, created, modified, sha1, sha256, md5, props) ❌
```

### After Fix:
```bash
jf rt search --spec=spec.json --include="name" --limit=1
# Returns: LIMITED fields (path, props only) ✅
```

## Related PR:
- jfrog-cli-core: https://github.com/jfrog/jfrog-cli-core/pull/1503

### DIFF
--- a/helm_test.go
+++ b/helm_test.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"testing"
+
+	coreTests "github.com/jfrog/jfrog-cli-core/v2/utils/tests"
+	"github.com/jfrog/jfrog-cli/utils/tests"
+)
+
+var (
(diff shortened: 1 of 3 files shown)


reshmifrog:
- please have a look why these tests are failing before merging

PR #3259, opened 2025-12-09
Bump version to 2.86.0

ehl-jf:
Bump version to 2.86.0


ehl-jf:
Transfer tests are failing because an infra issues `Cannot create project, Max number of projects is reached.`

ehl-jf:
The others tests need to be re-aligned.

PR #3260, opened 2025-12-09
Fix/lifecycle tests local artifactory

agrasth:

---


artifactory_test.go
```diff
@@ -6492,7 +6492,7 @@ func deleteProjectIfExists(t *testing.T, accessManager *access.AccessServicesMan
 	err := accessManager.DeleteProject(projectKey)
 	if err != nil {
 		if !strings.Contains(err.Error(), "Could not find project") {
-			t.Error(t, err)
```
agrasth, line 6495: `t.Error(t, err)` : Tries to print the testing object itself, then the error

PR #3276, opened 2025-12-15
Fix parsing Docker scan flags and Args

attiasas:

---

Depends on:
* https://github.com/jfrog/jfrog-cli-core/pull/1507

> Note: There is still an issue parsing flags/arguments.
> This PR only fixing that the inner docker scan cmd flags will be parsed as flags and not arguments.
> For  the scan to work you need to run: `docker scan  [flags]` (argements before flags)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -285,7 +285,7 @@ replace github.com/gfleury/go-bitbucket-v1 => github.com/gfleury/go-bitbucket-v1
 // replace github.com/jfrog/jfrog-cli-core/v2 => github.com/reshmifrog/jfrog-cli-core/v2 v2.58.5-0.20251110095124-ab9a4102ec5a
 
 // attiasas:convert_flags_public
-replace github.com/jfrog/jfrog-cli-core/v2 => github.com/attiasas/jfrog-cli-core/v2 v2.0.0-20251215112214-480d59f6e63e
+replace github.com/jfrog/jfrog-cli-core/v2 => github.com/attiasas/jfrog-cli-core/v2 v2.0.0-20251230074913-08e3d39269b0
 
 //replace github.com/jfrog/jfrog-client-go => github.com/jfrog/jfrog-client-go v1.54.2-0.20251007084958-5eeaa42c31a6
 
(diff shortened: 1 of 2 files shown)


buildtools/cli.go
```diff
@@ -494,6 +495,27 @@ func skipFlagParsingForDockerCmd() bool {
 	return true
 }
 
+func getDockerFlags() []cli.Flag {
```
RemiBou, line 500: no test ?

buildtools/cli.go
```diff
@@ -494,6 +495,27 @@ func skipFlagParsingForDockerCmd() bool {
...
+	// Adding `Docker Scan` flags
+	converted, _, err := components.ConvertFlags(securityDocs.DockerScan, securityDocs.GetCommandFlags(securityDocs.DockerScan))
+	if err != nil {
+		log.Error("Could not convert Docker Scan flags:", err)
+		return flagList
+	}
+	// Avoiding flag duplication
+	for _, f := range converted {
+		if !flagNames.Exists(f.GetName()) {
+			flagList = append(flagList, f)
```
RemiBou, line 516: why not use flagNames in the whole method and then convert it to slice ?

buildtools/cli.go
attiasas: added

buildtools/cli.go
attiasas: if we do it, when converting to slice you will need to go over the 2 sources of flags to locate where it is.
flagNames used only so we can validate quick if a flag already exist. I can switch it to just check if the slice contains the value and remove flagNames from the func... IMO its ok like this

ehl-jf:
@attiasas What about creating a new function like https://github.com/jfrog/jfrog-cli/blob/master/buildtools/cli.go#L1131 for example `extractDockerScanOptionsFromArgs` and adding additional options to `securityCLI.DockerScan` to provide the security's specific options ?

attiasas:
will be implemented at:
* https://github.com/jfrog/jfrog-cli-core/pull/1511

PR #3282, opened 2025-12-17
Supporting Run native behavior for Gradle

nitinp19:
We have enabled run native environmental variable which will let the user to run gradle native command. If the environmental variable is enabled, we will run the native flow.

### DIFF
--- a/buildtools/cli.go
+++ b/buildtools/cli.go
@@ -352,7 +352,7 @@ func GetCommands() []cli.Command {
 			UsageText:       helmcommand.GetArguments(),
 			ArgsUsage:       common.CreateEnvVars(),
 			SkipFlagParsing: true,
-			Hidden:          true,
+			Hidden:          false,
 			BashComplete:    corecommon.CreateBashCompletionFunc(),
 			Category:        buildToolsCategory,
 			Action:          HelmCmd,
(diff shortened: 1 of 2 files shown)


testdata/gradle/gradleproject/src/main/java/example/Example.java
```diff
@@ -0,0 +1,19 @@
+package example;
+
+/**
+ * Simple example class for the gradle test project.
+ * Note: We intentionally don't use JUnit here because dependency resolution
+ * may not work when the JFrog Gradle plugin configures the resolver to use
+ * the test Artifactory repository (which may not have JUnit cached).
+ */
```
bhanurp:
```suggestion

 // Simple example class for the gradle test project.
 // Note: We intentionally don't use JUnit here because dependency resolution
 // may not work when the JFrog Gradle plugin configures the resolver to use
 // the test Artifactory repository (which may not have JUnit cached).
```

buildtools/cli.go
```diff
@@ -631,10 +639,63 @@ func GradleCmd(c *cli.Context) (err error) {
...
+			if err != nil {
+				return cleanedArgs, nil, err
+			}
+			if serverDetails == nil {
+				return cleanedArgs, nil, fmt.Errorf("no default server configuration found. Please configure a server using 'jfrog config add' or specify a server using --server-id")
+			}
+			return cleanedArgs, serverDetails, nil
+		}
+
+		serverDetails, err := coreConfig.GetSpecificConfig(serverID, true, true)
```
naveenku-jfrog, line 659: This line should be before serverDetails, err := coreConfig.GetDefaultServerConf() condition. Extracting Default Server should be last option.

nitinp19: we are checking that with if block above

PR #3332, opened 2026-01-29
Added e2e tests for multiple package managers

bhanurp:

---
## Description
Added new e2e tests for multiple package managers to test CI VCS properties.


conan_test.go
```diff
@@ -467,3 +468,104 @@ func cleanupConanRemote() {
...
+			assert.Contains(t, props.Properties["vcs.org"], actualOrg, "Wrong vcs.org on %s", artifact.Name)
+
+			assert.Contains(t, props.Properties, "vcs.repo", "Missing vcs.repo on %s", artifact.Name)
+			assert.Contains(t, props.Properties["vcs.repo"], actualRepo, "Wrong vcs.repo on %s", artifact.Name)
+
+			artifactCount++
+		}
+	}
+
+	assert.Greater(t, artifactCount, 0, "No artifacts were validated for CI VCS properties")
```
agrasth, line 570: Comment says "No artifacts were validated" but message could be more specific about what went wrong.

nuget_test.go
```diff
@@ -475,3 +475,62 @@ func TestDotnetRequestedByDeterminism(t *testing.T) {
...
+	require.NotEmpty(t, publishedBuildInfo.BuildInfo.Modules, "Build info should have modules")
+	assert.Greater(t, len(publishedBuildInfo.BuildInfo.Modules[0].Dependencies), 0, "Build info should have dependencies")
+	assert.Equal(t, buildName, publishedBuildInfo.BuildInfo.Name, "Build name should match")
+	assert.Equal(t, buildNumber, publishedBuildInfo.BuildInfo.Number, "Build number should match")
+
+	// Verify CI environment was detected (for nuget push scenarios, CI VCS props would be set)
+	assert.NotEmpty(t, actualOrg, "CI org should be detected")
+	assert.NotEmpty(t, actualRepo, "CI repo should be detected")
+	t.Logf("NuGet build info created with %d dependencies in CI environment (org=%s, repo=%s)",
+		len(publishedBuildInfo.BuildInfo.Modules[0].Dependencies), actualOrg, actualRepo)
```
agrasth, line 535: This only checks if CI environment variables exist, but never validates that VCS properties were actually set on the artifacts in Artifactory. If needed we can add VCS property validation like other tests.
Or if NuGet doesn't support VCS properties on restore (only on push), we can rename the test.

pip_test.go
```diff
@@ -688,3 +688,92 @@ func TestSetupPipCommand(t *testing.T) {
...
+		for _, artifact := range module.Artifacts {
+			var fullPath string
+			switch {
+			case artifact.OriginalDeploymentRepo != "":
+				fullPath = artifact.OriginalDeploymentRepo + "/" + artifact.Path
+			case artifact.Path != "":
+				fullPath = artifact.Path
+			default:
+				continue // Skip artifacts without any path info
+			}
```
agrasth, line 758: All test files: artifactory_test.go, conan_test.go, go_test.go, helm_test.go, pip_test.go, poetry_test.go two code blocks are duplicated 6 times across all test files. We can consider creating a helper function, maybe in a new file test_helpers.go or add to existing test utilities.

helm_test.go
```diff
@@ -803,6 +805,159 @@ func TestHelmPackageAndPushWithBuildInfo(t *testing.T) {
...
+
+			// Validate VCS properties
+			assert.Contains(t, props.Properties, "vcs.provider", "Missing vcs.provider on %s", artifact.Name)
+			assert.Contains(t, props.Properties["vcs.provider"], "github", "Wrong vcs.provider on %s", artifact.Name)
+
+			assert.Contains(t, props.Properties, "vcs.org", "Missing vcs.org on %s", artifact.Name)
+			assert.Contains(t, props.Properties["vcs.org"], actualOrg, "Wrong vcs.org on %s", artifact.Name)
+
+			assert.Contains(t, props.Properties, "vcs.repo", "Missing vcs.repo on %s", artifact.Name)
+			assert.Contains(t, props.Properties["vcs.repo"], actualRepo, "Wrong vcs.repo on %s", artifact.Name)
```
agrasth, line 952: If needed we can also create a helper method for this, as these are also common in all files.

helm_test.go
```diff
@@ -803,6 +805,159 @@ func TestHelmPackageAndPushWithBuildInfo(t *testing.T) {
...
+				fullPath = artifact.Path
+			default:
+				continue // Skip artifacts without any path info
+			}
+
+			props, err := serviceManager.GetItemProps(fullPath)
+			assert.NoError(t, err, "Failed to get properties for artifact: %s", fullPath)
+			if props == nil {
+				continue
+			}
```
agrasth, line 942: When props is nil, test continues instead of failing. Is this expected? Or maybe we can use something like,
```
require.NotNil(t, props, "Properties should not be nil for artifact: %s (path: %s)", artifact.Name, fullPath)
```

artifactory_test.go
```diff
@@ -6885,3 +6885,80 @@ func validateSearchPatternWithAPI(t *testing.T, pattern string, recursive bool)
...
+
+// TestUploadMultipleFilesWithCIVcsProps tests that CI VCS properties are set on multiple uploaded files.
+// This test:
+// - Creates 15 files dynamically during the test
+// - Uploads all files using jf rt u
+// - Verifies vcs.provider, vcs.org, vcs.repo properties are set on ALL 15 files
+// - Fails if any property is missing on any file
+func TestUploadMultipleFilesWithCIVcsProps(t *testing.T) {
+	initArtifactoryTest(t, "")
+	const numFiles = 15
```
agrasth, line 6897: Extract Magic Numbers to Constants
```suggestion
// numFiles tests batch property setting - 15 is sufficient to verify
// properties are set on all files without making the test too slow
const numFiles = 15
```

conan_test.go
agrasth, line 570:
```suggestion
require.Greater(t, artifactCount, 0, 
	"No artifacts were validated for CI VCS properties. Build info had %d modules. Check if artifacts were actually uploaded.",
	len(publishedBuildInfo.BuildInfo.Modules))```

PR #3379, opened 2026-03-03
Added logic for updating HF_ENDPOINT and HF_TOKEN from default or spe…

naveenku-jfrog:

---
What: Added logic for updating HF_ENDPOINT and HF_TOKEN from default or specific server details itself
Why: Taking advantage of already stored config instead of manually exporting token and endpoint.
Testing: Done, on artifactory instance.


buildtools/cli.go
```diff
@@ -1235,6 +1240,53 @@ func huggingFaceDownloadCmd(c *cli.Context, hfArgs []string) error {
...
+}
+
+func updateHuggingFaceEnv(c *cli.Context, serverDetails *coreConfig.ServerDetails) error {
+	if os.Getenv(HF_ENDPOINT) == "" {
+		repoKey := c.String("repo-key")
+		if repoKey == "" {
+			return cliutils.PrintHelpAndReturnError("Please specify a repository key.", c)
+		}
+		hfEndpoint := serverDetails.GetArtifactoryUrl() + huggingfaceAPI + "/" + repoKey
+		err := os.Setenv(HF_ENDPOINT, hfEndpoint)
```
agrasth, line 1272: We can add some logs stating "Setting HF_ENDPOINT to:", hfEndpoint , WDYT?

agrasth (approved):

buildtools/cli.go
naveenku-jfrog: Actually we should not expose HF_ENDPOINT and HF_TOKEN in logs.

PR #3422, opened 2026-04-09
Jfrogpipelines

naveenku-jfrog:
---
Updated pipelines.yml and access.test so that these tests become compatible with tests running in GHE runners.

### DIFF
--- a/.github/workflows/helmTests.yml
+++ b/.github/workflows/helmTests.yml
@@ -1,6 +1,17 @@
 name: Helm Tests
 on:
   workflow_dispatch:
+    inputs:
+      jfrog_url:
+        description: "External JFrog Platform URL. Leave empty for local Artifactory."
+        type: string
+        required: false
+        default: ""
+      jfrog_admin_token:
+        description: "Admin token for external JFrog Platform."
+        type: string
+        required: false
+        default: ""
   push:
     branches:
       - "master"
@@ -81,6 +92,8 @@ jobs:
         uses: jfrog/.github/actions/install-local-artifactory@main
         with:
           RTLIC: ${{ secrets.RTLIC }}
+          JFROG_URL: ${{ inputs.jfrog_url }}
+          JFROG_ADMIN_TOKEN: ${{ inputs.jfrog_admin_token }}
           RT_CONNECTION_TIMEOUT_SECONDS: ${{ env.RT_CONNECTION_TIMEOUT_SECONDS || '1200' }}
 
       - name: Get ID Token and Exchange Token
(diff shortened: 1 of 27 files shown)


.github/workflows/helmTests.yml
```diff
@@ -1,6 +1,17 @@
...
+      jfrog_url:
+        description: "External JFrog Platform URL. Leave empty for local Artifactory."
+        type: string
+        required: false
+        default: ""
+      jfrog_admin_token:
+        description: "Admin token for external JFrog Platform."
+        type: string
+        required: false
+        default: ""
```
bhanurp, line 14: this looks like repeating pattern fr all the files can it be made generic?

.github/workflows/pluginsTests.yml
```diff
@@ -69,11 +80,15 @@ jobs:
...
+          JFROG_URL: ${{ inputs.jfrog_url }}
+          JFROG_ADMIN_TOKEN: ${{ inputs.jfrog_admin_token }}
           RT_CONNECTION_TIMEOUT_SECONDS: ${{ env.RT_CONNECTION_TIMEOUT_SECONDS || '1200' }}
 
       - name: Run plugins tests
         if: matrix.os.name != 'macos'
-        run: go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.plugins
+        run: >-
+          go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.plugins
+          ${{ env.JFROG_TESTS_IS_EXTERNAL == 'true' && format('--jfrog.url={0} --jfrog.adminToken={1}', env.JFROG_TESTS_URL, env.JFROG_TESTS_LOCAL_ACCESS_TOKEN) || '' }}
```
bhanurp, line 94: since token is used here, does tests leak the token used fr tests? were you able to verify?

.github/workflows/helmTests.yml
naveenku-jfrog: GitHub Actions has no mechanism to share or inherit workflow_dispatch input definitions across workflow files. Each file must declare its own inputs independently — this is a documented platform limitation, not a design choice we can easily escape.

.github/workflows/pluginsTests.yml
naveenku-jfrog: This is the already used style in all workflows. Refer: https://github.com/jfrog/jfrog-cli/blob/master/.github/workflows/artifactoryTests.yml#L82

bhanurp (approved):

agrasth (approved):

PR #3438, opened 2026-04-15
Feature/rteco 1003 external artifactory support

naveenku-jfrog:
---
**Composite Action: install-local-artifactory/action.yml**
What: Added two new optional inputs (JFROG_URL, JFROG_ADMIN_TOKEN) to the composite action. When JFROG_URL is provided, the action skips the local Artifactory installation entirely and instead exports three environment variables: JFROG_TESTS_LOCAL_ACCESS_TOKEN, JFROG_TESTS_URL, and JFROG_TESTS_IS_EXTERNAL. When JFROG_URL is empty (default), the existing local-rt-setup logic runs unchanged and JFROG_TESTS_URL is set to http://127.0.0.1:8082.

Why: Today, every test workflow installs a fresh local Artifactory instance via local-rt-setup. This works for open-source CI on github.com but blocks us from running the same tests against a pipeline-provisioned draft Artifactory instance (for validating pre-release builds). By making the Artifactory source configurable, we can reuse the exact same workflows for both local and external instances without duplicating workflow files.

**workflow_dispatch Inputs (all 14 workflow files)**
What: Added jfrog_url and jfrog_admin_token as optional workflow_dispatch inputs with empty string defaults to all 14 test workflows.

Why: These inputs allow an external caller (JFrog Pipelines, GHE dispatch, or manual trigger) to pass in an external Artifactory URL and admin token. When triggered by push or pull_request_target, these inputs are empty and the workflow behaves identically to before — zero impact on existing CI.

**Pattern A Workflows: Replace Hardcoded URL (6 workflows)**
Files: artifactoryTests.yml, accessTests.yml, helmTests.yml, huggingfaceTests.yml, lifecycleTests.yml, transferTests.yml (Transfer-7 job)

What: Replaced the hardcoded --jfrog.url=http://127.0.0.1:8082 with --jfrog.url=${{ env.JFROG_TESTS_URL || 'http://127.0.0.1:8082' }} in the go test commands.

Why: These workflows had the local Artifactory address baked into the test command. By reading from the JFROG_TESTS_URL environment variable (set by the composite action), the same go test command works for both local and external Artifactory. The || 'http://127.0.0.1:8082' fallback ensures backward compatibility even if the composite action hasn't been updated yet.

**Pattern B Workflows: Add Conditional Flags (8 workflows + 2 special cases)**
Files: npmTests.yml, mavenTests.yml, gradleTests.yml, conanTests.yml, pnpmTests.yml, nugetTests.yml, pythonTests.yml, pluginsTests.yml, plus artifactoryTests.yml (artifactoryProject step) and transferTests.yml (Transfer-6 job)

What: Added a conditional expression that appends --jfrog.url and --jfrog.adminToken flags only when JFROG_TESTS_IS_EXTERNAL == 'true':

${{ env.JFROG_TESTS_IS_EXTERNAL == 'true' && format('--jfrog.url={0} --jfrog.adminToken={1}', env.JFROG_TESTS_URL, env.JFROG_TESTS_LOCAL_ACCESS_TOKEN) || '' }}
Why: These workflows never passed --jfrog.url or --jfrog.adminToken — they relied on Go flag defaults (http://localhost:8081/) and the GetLocalArtifactoryTokenIfNeeded() function, which reads the token from the JFROG_TESTS_LOCAL_ACCESS_TOKEN env var only if the URL contains localhost:808. For an external Artifactory, those defaults don't work. However, unconditionally adding the flags would change the URL from port 8081 to 8082 for local installs, which could introduce subtle differences. The conditional approach preserves exact existing behavior for local installs (no flags appended, original defaults used) while adding the required flags only for external instances.

**Transfer-7: Added Explicit --jfrog.adminToken**
What: Added --jfrog.adminToken=${{ env.JFROG_TESTS_LOCAL_ACCESS_TOKEN }} to the Transfer-7 go test command (previously absent).

Why: Transfer-7 previously relied on the Go flag default mechanism to pick up the token — it passed --jfrog.url=http://127.0.0.1:8082 but not --jfrog.adminToken. The token default works because at flag registration time, the default URL is localhost:8081 (which matches GetLocalArtifactoryTokenIfNeeded). With an external URL, this implicit token resolution breaks. Making it explicit is backward-compatible (same env var value) and necessary for external instances.

### DIFF
--- a/.github/workflows/accessTests.yml
+++ b/.github/workflows/accessTests.yml
@@ -1,6 +1,17 @@
 name: Access Tests
 on:
   workflow_dispatch:
+    inputs:
+      jfrog_url:
+        description: "External JFrog Platform URL. Leave empty for local Artifactory."
+        type: string
+        required: false
+        default: ""
+      jfrog_admin_token:
+        description: "Admin token for external JFrog Platform."
+        type: string
+        required: false
+        default: ""
   push:
     branches:
       - "master"
@@ -75,6 +86,8 @@ jobs:
         uses: jfrog/.github/actions/install-local-artifactory@main
         with:
           RTLIC: ${{ secrets.RTLIC }}
+          JFROG_URL: ${{ inputs.jfrog_url }}
+          JFROG_ADMIN_TOKEN: ${{ inputs.jfrog_admin_token }}
           RT_CONNECTION_TIMEOUT_SECONDS: ${{ env.RT_CONNECTION_TIMEOUT_SECONDS || '1200' }}
 
       - name: Get ID Token and Exchange Token
@@ -87,4 +100,7 @@ jobs:
 
       - name: Run Access tests
         if: matrix.os.name != 'macos'
-        run: go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.access --jfrog.url=http://127.0.0.1:8082 --jfrog.adminToken=${{ env.JFROG_TESTS_LOCAL_ACCESS_TOKEN }}
+        run: >-
+          go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.access
+          --jfrog.url=${{ env.JFROG_TESTS_URL || 'http://127.0.0.1:8082' }}
+          --jfrog.adminToken=${{ env.JFROG_TESTS_LOCAL_ACCESS_TOKEN }}
--- a/.github/actions/install-local-artifactory/action.yml
+++ b/.github/actions/install-local-artifactory/action.yml
@@ -1,0 +1,55 @@
+name: "Setup Local Artifactory"
+description: "Setup Local Artifactory or connect to an external JFrog Platform instance"
+inputs:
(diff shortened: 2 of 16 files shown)


.github/workflows/accessTests.yml
```diff
@@ -87,4 +100,7 @@ jobs:
 
       - name: Run Access tests
         if: matrix.os.name != 'macos'
-        run: go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.access --jfrog.url=http://127.0.0.1:8082 --jfrog.adminToken=${{ env.JFROG_TESTS_LOCAL_ACCESS_TOKEN }}
+        run: >-
```
bhanurp, line 103: does this change impact the integration tests run on jfrog-cli-artifactory?

naveenku-jfrog: No, all the changes are backward compatible.

PR #3440, opened 2026-04-16
RTECO-1035 - Added Support for yarn v4

fluxxBot:

---


npm_test.go
```diff
@@ -1119,12 +1126,40 @@ func TestYarnInV4(t *testing.T) {
...
+	assert.NoError(t, yarn.ConfigSet("unsafeHttpWhitelist", "[\"localhost\"]", yarnExecPath, true))
+	defer func() {
+		assert.NoError(t, yarn.ConfigSet("unsafeHttpWhitelist", origWhitelist, yarnExecPath, true))
+	}()
+
 	jfrogCli := coretests.NewJfrogCli(execMain, "jfrog", "")
-	err = jfrogCli.Exec("yarn", "install")
-	assert.Error(t, err)
+	// Yarn v4 is now supported — install should succeed and collect build-info
+	assert.NoError(t, jfrogCli.Exec("yarn", "--build-name="+tests.YarnBuildName, "--build-number=2", "--module=yarnV4Module"))
```
bhanurp, line 1194: are the e2e scenarios for v4 support added, especially new features which are added as part of v4?

fluxxBot: A few things worth noting about how build-info collection works here, which I think addresses the underlying concern:

1. We don't tamper with the yarn install process itself. Yarn runs normally (native mode just uses the user's own .yarnrc.yml; configured mode swaps in our registry/auth and restores afterwards). What yarn resolves and installs is 100% yarn's output — we don't modify the lockfile, the cache, or any resolver behaviour.

2. Dependency capture is delegated to yarn. We collect the tree via yarn info --all --recursive --json — the same command build-info-go has used for every yarn 2+ version. Whatever yarn reports as a dependency (npm, workspace, patch, portal, file, git, exec, plugins, etc.) is what we record. New v4 protocols or tree shapes automatically flow through without special handling. 

3. Checksums come from Artifactory via AQL, not from local parsing. For every dep yarn emits, we ask Artifactory "do you have this name:version? what are its checksums?" — a standard flow shared with npm/pnpm. Anything Artifactory knows about gets recorded; anything it doesn't is surfaced in the "missing dependencies" warning, not failed.

4. Parsing is version-agnostic. The JSON shape yarn emits (value, children.Version, children.Dependencies with descriptor/locator) is identical between v2, v3, and v4. We verified this by running the same project through both v3.6.4 and v4.5.0 — output is byte-for-byte equivalent.

bhanurp: can you verify below scenarios work?
  1. Workspace monorepo root — build-info empty even though packages are installed.
  2. Inside a workspace package — workspace siblings (workspace:*) and their transitive deps dropped from build-info.
  3. link: protocol — local linked package dropped, misleading "could not be found in Artifactory" warning.
  4. file: protocol — local tarball dep dropped from build-info.
  5. portal: protocol — local portal dep dropped from build-info.
  6. git+https:// dep — git-sourced dep dropped from build-info.
  7. patch: protocol — build-info records the unpatched tarball's checksum, not the actual patched code on disk.

bhanurp (approved):

Issue #2309, opened 2023-11-13
Unable to use private PKI certificate


fourpastmidnight:
### Describe the bug

We use an internal/private PKI. We are attempting to use the JFrog Azure DevOps Server marketplace add-in, which under the covers uses this Jfrog CLI.

The documentation states that when using a private PKI, one must make the certificate(s) available to the Jfrog CLI because the Jfrog CLI is currently incapable of using the Windows Certificate Store (for Windows platforms). According to the documentation, the certificate must be located at `~/.jfrog/security/certs` (though, I have seen other documentation suggest the location `~/.jfrog/security`).

For the purposes of the Azure DevOps Extension, the actual location of the Jfrog CLI (e.g. the Home directory) is `$(Agent.ToolsDirectory)/_jf`.

In an attempt to get a working build pipeline, the following tests were performed:

1. Using the Jfrog DotNet Core task:
    1. Place the full certificate chain, including the Artifactory server certificate, into `$(Agent.ToolsDirectory)/_jf/security/certs` in a file named `custom.crt` in PEM format
    2. Place only the chain of trust (so, excluding the Artifactory server certificate) into the same file in the same location as above
    3. Repeat tests 1 and 2 above, but this time using the path `$(Agent.ToolsDirectory)/_jf/security`.
    4. Setting the `JFROG_CLI_HOME` environment variable to `$(Agent.ToolsDirectory)/_jf` since it appears from the source code of the Azure DevOps Server add-in task that this environment variable is not set by that task, therefore resulting in the "default location" of the Jfrog CLI home directory being used, causing the certificate to potentially not be found. (A bug issue has been filed on the Jfrog Azure DevOps Server extension repository)
    5. Repeating test 4 above, but appending a `/` to the environment variable value.
2. Using the Jfrog CLI V2 task:
    1. Recreate the command executed by the Jfrog DotNet Core task, and re-execute all 4 tests above under &sect;1
    2. Add the `--insecure-tls` argument to the command in test 1 immediately above. **NOTE: This is the only test that succeeded.**

[...]

PR #3441, opened 2026-04-16
Automatic import of private CA certificates in Windows.

Cesar-DC:
## Summary

This PR adds a Windows-specific preflight step to improve handling of private PKI certificates in JFrog CLI.

When a command includes a target URL, the CLI:

1. Performs a TLS handshake using the Windows trust store
2. Extracts the verified certificate chain
3. Writes the certificates (leaf, intermediate, root) into:
   ~/.jfrog/security/certs
4. Continues execution

## Motivation

On Windows, JFrog CLI may fail with:

even when the certificate is already trusted by the OS.

Manually exporting the site certificate from a browser and placing it in `~/.jfrog/security/certs` resolves the issue.

This patch automates that process.

## Behavior

- Windows only
- Triggered when `--url` or `JFROG_URL` is provided
- Writes:
  - leaf.pem
  - intermediate.pem
  - root.pem

## Testing

Validated locally:

```powershell
jf.exe rt ping --url=https://<host>/artifactory

### DIFF
--- a/internal/certbootstrap/persist.go
+++ b/internal/certbootstrap/persist.go
@@ -1,0 +1,71 @@
+package certbootstrap
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func ResolveJfrogHomeDir() string {
+	if v := os.Getenv("JFROG_CLI_HOME_DIR"); v != "" {
+		return v
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return ".jfrog"
+	}
+	return filepath.Join(home, ".jfrog")
+}
+
+func PersistVerifiedChain(jfrogHome string, chain []*x509.Certificate) (int, error) {
+	certsDir := filepath.Join(jfrogHome, "security", "certs")
+	if err := os.MkdirAll(certsDir, 0o700); err != nil {
+		return 0, err
+	}
+
+	fmt.Fprintf(os.Stderr, "[certbootstrap] persisting to cert dir=%s\n", certsDir)
+
+	written := 0
+
+	for i, cert := range chain {
+		fmt.Fprintf(
+			os.Stderr,
+			"[certbootstrap] persist input chain[%d] subject=%s issuer=%s isCA=%v\n",
+			i,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: internal/certbootstrap/persist.go [hunk 1]
```diff
@@ -0,0 +1,71 @@
+package certbootstrap
```
RemiBou, line 1: add tests

internal/certbootstrap/persist.go
```diff
@@ -0,0 +1,71 @@
...
+		return v
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return ".jfrog"
+	}
+	return filepath.Join(home, ".jfrog")
+}
+
+func PersistVerifiedChain(jfrogHome string, chain []*x509.Certificate) (int, error) {
```
RemiBou, line 22: why do we need to persist the chain ?

RemiBou (changes requested):
add tests

internal/certbootstrap/persist.go
Cesar-DC: Go on Windows can validate a certificate against the CAs available in the Windows certificate store (typically managed by a corporate PKI team).

If the connection is successfully validated by Go, we can safely extract and copy the certificate chain into the location expected by JFrog CLI (~/.jfrog/security/certs).

This approach avoids a full rewrite of the CLI to directly support the Windows certificate store. Instead, it preserves the existing mechanism (manual placement of private CA certificates), while leveraging Go’s ability to interact with the system certificate store to ensure the chain we import is trusted.

RemiBou:
It should be handled in https://github.com/jfrog/jfrog-client-go/blob/master/auth/cert/sslutils_windows.go

Cesar-DC:
Even better, we have a 5 lines fix then

https://github.com/jfrog/jfrog-client-go/blob/master/auth/cert/loader.go


replace:
caCertPool, err := loadSystemRoots()

with:
caCertPool, err := x509.SystemCertPool()
if caCertPool == nil {
	caCertPool = x509.NewCertPool()
}

Tested locally. 
might be a fix for : https://github.com/jfrog/jfrog-cli/issues/2309

Issue #1060, opened 2021-04-20
jfrog.exe rt download can crash when --min-split is smaller than the largest ZIP file


RudiMT:
After jfrog.exe 1.46.2 fixed most of #1002, only remaining corner case is when --spec-vars=D:/my/path/ *with a trailing (forward) slash*  is used, producing a `//` in the file, and --min-split=5120 is *smaller* than the largest of the ZIP files. This cannot always be controlled by the end user.

`jfrog1.46.2-dev.exe rt dl --spec=D:/try001/repro.json --spec-vars="location=d:/try001/" --min-split=5120`
```
[Debug] Replacing '${location}' with 'd:/try001/'
[Debug] The reformatted content is:
{
  "files": [
    {
      "pattern": "generic-local/02-Externals/01-Libs/FileSize10000k.zip",
      "target": "d:/try001//02-Externals/01-Libs/",
      "flat": "true",
      "explode": "true"
    },
...
```

Setting --min-split=-1 (or --min-split=11000, which is larger than file FileSize10000k.zip in the example) works in this case.

_Originally posted by @RudiMT in https://github.com/jfrog/jfrog-cli/issues/1002#issuecomment-820960267_

RudiMT:
#1002, new issue as suggested by @Mindavi

asafgabai:
Hi @RudiMT,
Thanks for reporting this issue.
I opened a PR for fixing this [here](https://github.com/jfrog/jfrog-client-go/pull/399), and it'll be released soon.

asafgabai:
@RudiMT
JFrog CLI 2.2.0 and 1.51.0 are released and include this fix.
We'd appreciate your feedback for it.

RudiMT:
We looked at 1.51.0 and it works.

PR #3442, opened 2026-04-17
RTECO-1049 - switch between podman and docker

fluxxBot:

---

### DIFF
--- a/buildtools/cli.go
+++ b/buildtools/cli.go
@@ -1045,14 +1045,30 @@ func goCmdVerification(c *cli.Context) (string, error) {
 }
 
 // resolveContainerManagerType returns the container manager to use when running 'jf docker' subcommands.
-// It defaults to Docker but can be overridden to Podman via the JFROG_CLI_CONTAINER_MANAGER env var.
-// This allows users running Podman (or the podman-docker shim) to keep using 'jf docker ...' in their
-// existing pipelines without switching to 'jf rt podman-*'.
+//
+// If the local 'docker' binary reports Podman in its version output (i.e. the podman-docker shim
+// or native podman aliased as docker), treat it as Podman so 'jf docker ...' works transparently
+// for Podman users without daemon-socket access. Otherwise default to Docker.
+//
+// Detection is intentionally conservative: only a positive "Podman" signal from 'docker version'
+// switches behavior. Real Docker installations are unaffected.
 func resolveContainerManagerType() containerutils.ContainerManagerType {
-	if strings.EqualFold(os.Getenv("JFROG_CLI_CONTAINER_MANAGER"), "podman") {
+	if isPodmanBackedDockerCli() {
+		log.Debug("Detected Podman-backed 'docker' CLI. Routing 'jf docker' subcommands through Podman.")
 		return containerutils.Podman
 	}
 	return containerutils.DockerClient
+}
+
+// isPodmanBackedDockerCli returns true if the local 'docker' binary is actually Podman
+// (either via the podman-docker shim or an alias). Any error or missing binary returns false.
+func isPodmanBackedDockerCli() bool {
+	cmd := exec.Command("docker", "version")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return false
+	}
+	return strings.Contains(strings.ToLower(string(out)), "podman")
 }
 
 func dockerCmd(c *cli.Context) error {
--- a/buildtools/cli.go
+++ b/buildtools/cli.go
@@ -1044,6 +1044,10 @@ func goCmdVerification(c *cli.Context) (string, error) {
 	return configFilePath, nil
 }
 
+// podmanDetector is indirected through a package-level variable so tests can
+// replace the real 'docker version' probe with a deterministic stub.
+var podmanDetector = isPodmanBackedDockerCli
+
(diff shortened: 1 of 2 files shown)


buildtools/cli.go
```diff
@@ -1058,6 +1044,17 @@ func goCmdVerification(c *cli.Context) (string, error) {
 	return configFilePath, nil
 }
 
+// resolveContainerManagerType returns the container manager to use when running 'jf docker' subcommands.
+// It defaults to Docker but can be overridden to Podman via the JFROG_CLI_CONTAINER_MANAGER env var.
+// This allows users running Podman (or the podman-docker shim) to keep using 'jf docker ...' in their
+// existing pipelines without switching to 'jf rt podman-*'.
+func resolveContainerManagerType() containerutils.ContainerManagerType {
+	if strings.EqualFold(os.Getenv("JFROG_CLI_CONTAINER_MANAGER"), "podman") {
```
itsmeleela: If someone sets JFROG_CLI_CONTAINER_MANAGER=podmann (with typo) or =docker, it silently falls back to Docker with no warning. A log message like "unrecognized JFROG_CLI_CONTAINER_MANAGER value '%s', defaulting to Docker" would save a lot of future debugging isnt it?

itsmeleela: Shouldn't these functions to have unit tests?

fluxxBot: not required now

fluxxBot: added

buildtools/cli.go
```diff
@@ -1058,6 +1044,37 @@ func goCmdVerification(c *cli.Context) (string, error) {
...
+		log.Debug("Detected Podman-backed 'docker' CLI. Routing 'jf docker' subcommands through Podman.")
+		return containerutils.Podman
+	}
+	return containerutils.DockerClient
+}
+
+// isPodmanBackedDockerCli returns true if the local 'docker' binary is actually Podman
+// (either via the podman-docker shim or an alias). Any error or missing binary returns false.
+func isPodmanBackedDockerCli() bool {
+	cmd := exec.Command("docker", "version")
```
bhanurp, line 1099: what will happen when docker version fails with unable to find executable?
isPodmanBackedDockerCli() -> returns false -> podmanDetector() returns false at ln #1060 returns `return containerutils.DockerClient`

buildtools/cli.go
```diff
@@ -1058,6 +1044,37 @@ func goCmdVerification(c *cli.Context) (string, error) {
...
+	if podmanDetector() {
+		log.Debug("Detected Podman-backed 'docker' CLI. Routing 'jf docker' subcommands through Podman.")
+		return containerutils.Podman
+	}
+	return containerutils.DockerClient
+}
+
+// isPodmanBackedDockerCli returns true if the local 'docker' binary is actually Podman
+// (either via the podman-docker shim or an alias). Any error or missing binary returns false.
+func isPodmanBackedDockerCli() bool {
```
bhanurp: nit pick:
```suggestion
func dockerIsPodman() bool {
```

buildtools/cli.go
```diff
@@ -1058,6 +1044,37 @@ func goCmdVerification(c *cli.Context) (string, error) {
...
+//
+// Detection is intentionally conservative: only a positive "Podman" signal from 'docker version'
+// switches behavior. Real Docker installations are unaffected.
+func resolveContainerManagerType() containerutils.ContainerManagerType {
+	if podmanDetector() {
+		log.Debug("Detected Podman-backed 'docker' CLI. Routing 'jf docker' subcommands through Podman.")
+		return containerutils.Podman
+	}
+	return containerutils.DockerClient
+}
```
bhanurp, line 1094: lets improve this incase users want to force the container manager
with an env like `JFROG_CONTAINER_MANAGER=docker|podman`

```suggestion
func resolveContainerManagerType() containerutils.ContainerManagerType {
    switch strings.ToLower(strings.TrimSpace(os.Getenv(containerManagerEnvVar))
        case "podman":
                log.Debug(containerManagerEnvVar + "=podman. Routing 'jf docker' su
  through Podman.")
                return containerutils.Podman
        case "docker":
                log.Debug(containerManagerEnvVar + "=docker. Routing 'jf docker' su
  through Docker.")
                return containerutils.DockerClient
        }
    
	if podmanDetector() {
		log.Debug("Detected Podman-backed 'docker' CLI. Routing 'jf docker' subcommands through Podman.")
		return containerutils.Podman
	}
	return containerutils.DockerClient
}
```

bhanurp (changes requested):

buildtools/cli.go
fluxxBot: When docker version fails (binary not installed, permission denied, etc.), the detector returns false and we fall through to the default DockerClient — identical to the pre-change behavior where jf docker was always hardcoded to DockerClient. So we don't regress anything; we only add a positive detection path for the podman-docker shim. Users with no docker binary at all were broken before and will still be broken (they need to install either real Docker or the podman-docker package)

buildtools/cli.go
fluxxBot: done!

buildtools/cli.go
fluxxBot: done

bhanurp (approved):

PR #3449, opened 2026-04-21
fix: build+pattern downloads against virtual repos

bhanurp:
## Summary
- Adds a `replace` directive in `go.mod` pointing `github.com/jfrog/jfrog-client-go` at the `fix/build-with-pattern-virtual-repo` branch so CI exercises the virtual-repo pattern-download fix.
- Depends on jfrog/jfrog-client-go#1338 — the replace should be removed once that PR merges and a client-go release is bumped here.

## Test plan
- [ ] CI green on this branch
- [ ] Manual: `jf rt download --build=<build>/<number> "<virtual-repo>/**"` returns the expected artifacts


artifactory_test.go
```diff
@@ -3502,6 +3502,126 @@ func TestArtifactoryDownloadByBuildUsingSimpleDownload(t *testing.T) {
...
 
+// TestArtifactoryDownloadByPatternAndBuildFromVirtualRepo guards the fix in jfrog-client-go
+// PR #1338: "build+pattern downloads against virtual repos". The client used to post-filter
+// build search results by comparing the pattern's repo against each result's Repo field.
+// AQL returns the backing local repo name, not the virtual repo the user typed, so any
+// download that combined --build with a pattern pointing at a virtual repo returned zero
+// files. The fix moves the pattern into the AQL query itself so Artifactory resolves the
+// virtual repo server-side. This test fails with "nothing downloaded" against the pre-fix
+// client-go.
+func TestArtifactoryDownloadByPatternAndBuildFromVirtualRepo(t *testing.T) {
```
fluxxBot, line 3513: should we add a test for local repo as well?

artifactory_test.go
```diff
@@ -3502,6 +3502,126 @@ func TestArtifactoryDownloadByBuildUsingSimpleDownload(t *testing.T) {
...
+	initArtifactoryTest(t, "")
+	buildNumberA, buildNumberB := "10", "11"
```
fluxxBot: there are runtime constants from which build number can be taken from

fluxxBot (approved):

artifactory_test.go
bhanurp: TestArtifactoryDownloadByPatternAndBuildUsingSpec, TestArtifactoryDownloadByBuildUsingSimpleDownload already takes care of local repository

artifactory_test.go
bhanurp: Good catch — applied in ab05106f. Added a `uniqueBuildNumberPair()` helper that returns timestamp-based build numbers so a dirty server from a crashed previous run can't collide with the current one. The spec-file variant now generates its spec inline too so the build number stays in sync without a fixture.

artifactory_test.go
bhanurp: Local-repo coverage for `--build` + pattern already exists: `TestArtifactoryDownloadByPatternAndBuildUsingSpec` (artifactory_test.go:3194) and `TestArtifactoryDownloadByBuildUsingSimpleDownload` (artifactory_test.go:3474) both use `tests.RtRepo1` (local). These three new tests exist specifically to guard the virtual-repo regression — adding a local-repo variant would duplicate the existing coverage.

PR #3462, opened 2026-04-29
Fix UV and Maven test failures

agrasth:

---

### DIFF
--- a/buildtools/cli.go
+++ b/buildtools/cli.go
@@ -1835,18 +1835,10 @@ func UvCmd(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	// urfave/cli v1 reorders defined flags to the front and parses them before UvCmd
-	// runs, so they are absent from c.Args(). Fall back to the parsed flag values.
-	if m := c.String("module"); m != "" {
-		buildConfiguration.SetModule(m)
-	}
 	var serverID string
 	filteredArgs, serverID, err = coreutils.ExtractServerIdFromCommand(filteredArgs)
 	if err != nil {
 		return fmt.Errorf("failed to extract server ID: %w", err)
-	}
-	if serverID == "" {
-		serverID = c.String("server-id")
 	}
 	cmdName, uvArgs := getCommandName(filteredArgs)
 	// Peek at --publish-url to populate DeployerRepo for build-info enrichment.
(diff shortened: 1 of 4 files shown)


buildtools/cli.go
```diff
@@ -1835,11 +1835,19 @@ func UvCmd(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+	// urfave/cli v1 reorders defined flags to the front and parses them before UvCmd
```
bhanurp: why is this change needed

utils/tests/proxy/server/server.go
```diff
@@ -21,6 +21,12 @@ type httpResponse func(rw http.ResponseWriter, req *http.Request)
 
 func handleReverseProxyHttps(reverseProxy *httputil.ReverseProxy) httpResponse {
 	return func(rw http.ResponseWriter, req *http.Request) {
+		// Health-check path used by checkIfServerIsUp — return 200 directly,
```
bhanurp, line 24: this proxy server is used only in testing?

buildtools/cli.go
```diff
@@ -1835,11 +1835,19 @@ func UvCmd(c *cli.Context) error {
...
+	// runs, so they are absent from c.Args(). Fall back to the parsed flag values.
+	if m := c.String("module"); m != "" {
+		buildConfiguration.SetModule(m)
+	}
 	var serverID string
 	filteredArgs, serverID, err = coreutils.ExtractServerIdFromCommand(filteredArgs)
 	if err != nil {
 		return fmt.Errorf("failed to extract server ID: %w", err)
 	}
+	if serverID == "" {
```
bhanurp: isnt serverID read in previous step why is extra check required and why is it set again by reading from flagvalue?

utils/tests/proxy/server/server.go
agrasth: Yes, that proxy server is test infrastructure only.

buildtools/cli.go
agrasth: You are right, it's not needed.  When `TestUvModuleOverrideOnPublish` and `TestUvServerIDFlag` were failing, I assumed urfave/cli v1 was consuming `--module` and `--server-id` before UvCmd saw them — based on the reorderArgs behavior. So I added `c.String("module")` and `c.String("server-id")` as fallbacks.

But `SkipFlagParsing: true` was already on the UV command from the original PR. With that set, urfave/cli skips ALL flag parsing, which dumps everything straight into c.Args(). So c.String() always returns "", and the fallbacks never executed. 
   They were dead code. I have removed it

bhanurp (approved):

Issue #1345, opened 2021-12-16
jfrog rt repo-template / repo-create (repo-update) should be project-aware


tsuyo:
**Is your feature request related to a problem? Please describe.**
There is no option to create a repo under a project

**Describe the solution you'd like to see**
Either repo-template or repo-create (repo-update) has a chance to specify its project

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #3493, opened 2026-05-18
Enable JSON error output for commands with HTTP requests

ehl-jf:
### Summary

Adds opt-in JSON output for HTTP error responses across the entire CLI. When enabled, any command that hits the platform — Artifactory, Access, Xray, Pipelines, OIDC token exchange — emits a structured JSON object on **stdout** (the data channel) instead of human-readable text on stderr. Default behavior is unchanged; this is purely additive.

### How to use

Two equivalent triggers:

| Mechanism | Example |
|---|---|
| Environment variable | `JFROG_CLI_ERROR_OUTPUT_FORMAT=json jf rt ping ...` |
| `--format=json` auto-promote | `jf rt ping --format=json ...` |

If both are set, the env var wins. Any value other than `json` (or unset) keeps the current text behavior.

### Channel contract (JSON mode)

| Stream | Contents |
|---|---|
| stdout | parseable JSON: `{status_code, status, body}` — body is a nested object when the server returned JSON |
| stderr | log lines only (`[Info]`/`[Warn]`/Trace ID), unchanged from today |
| exit | 1 |

Makes `jf <cmd> --format=json 1>out.json 2>err.log` work cleanly: `jq . out.json` parses, `err.log` is for humans.

### Coverage

- All commands routing through `jfrog-client-go`'s `CheckResponseStatus*` (i.e. the standard HTTP error path on every JFrog service).
- OIDC token exchange (`jf eot`) end-to-end.
- `jf api`: in JSON mode, the structured error replaces the body-on-stdout dump; in default mode, curl-like behavior is preserved.
- Transport errors (connection refused, TLS) intentionally fall through to the existing text path — there's no HTTP response to wrap.

### Companion PR

Depends on https://github.com/jfrog/jfrog-client-go/pull/1345, which exposes the typed `HttpResponseError` this PR consumes. go.mod is pinned to that branch's pseudo-version; bump to a released version once #1345 merges.

### Tests

Unit tests in `utils/cliutils/errorformat_test.go` (env var precedence, flag auto-promote variants, false-positive guards, legacy-text fallback) and `general/api/cli_test.go` (JSON-mode emits structured stdout, success path unchanged, default mode still dumps body).

---

---

### DIFF
--- a/utils/cliutils/errorformat.go
+++ b/utils/cliutils/errorformat.go
@@ -133,13 +133,15 @@ func parseLegacyHTTPResponseError(msg string) *errorutils.HttpResponseError {
 	statusLine = strings.TrimSpace(statusLine)
 
 	// First token must be a numeric status code (e.g. "401 Unauthorized").
-	// Cap the digit count so Atoi cannot overflow on malformed input — real
-	// HTTP statuses are 3 digits; 9 leaves plenty of headroom under int32.
+	// Real HTTP statuses are 3 digits. We accept up to 9 to leave headroom
+	// under int32, but anything longer is malformed — return nil so the
+	// caller falls back to the text path rather than emitting a truncated
+	// status_code the server never sent.
 	codeEnd := 0
-	for codeEnd < len(statusLine) && codeEnd < 9 && statusLine[codeEnd] >= '0' && statusLine[codeEnd] <= '9' {
+	for codeEnd < len(statusLine) && statusLine[codeEnd] >= '0' && statusLine[codeEnd] <= '9' {
 		codeEnd++
 	}
-	if codeEnd == 0 {
+	if codeEnd == 0 || codeEnd > 9 {
 		return nil
 	}
 	code, _ := strconv.Atoi(statusLine[:codeEnd])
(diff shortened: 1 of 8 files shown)


utils/cliutils/errorformat.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	rest := msg[idx+len(legacyHTTPErrPrefix):]
+	statusLine, bodyStr, _ := strings.Cut(rest, "\n")
+	statusLine = strings.TrimSpace(statusLine)
+
+	// First token must be a numeric status code (e.g. "401 Unauthorized").
+	// Cap the digit count so Atoi cannot overflow on malformed input — real
+	// HTTP statuses are 3 digits; 9 leaves plenty of headroom under int32.
+	codeEnd := 0
+	for codeEnd < len(statusLine) && codeEnd < 9 && statusLine[codeEnd] >= '0' && statusLine[codeEnd] <= '9' {
```
vitaliil-jfrog: Do we have a test case for a long error code, more than 9 digits?

vitaliil-jfrog (approved):

christophermiJfrog (approved):

utils/cliutils/errorformat.go
ehl-jf: Good catch — we didn't. Pushed `aa089ab5`.

PR #3497, opened 2026-05-20
Bump version to 2.104.1

christophermiJfrog:
Bump version to 2.104.1


.github/scripts/prepare-cli-release
```diff
@@ -117,7 +117,7 @@ def compute_next_version(version_type: str, gh_token: Optional[str] = None) -> s
         sys.exit(1)
     
     print(f"Next version: {next_version}")
-    return next_version
+    return "2.104.1"
```
ehl-jf, line 120: Why this change ?
If done by the script it should be fixed

PR #3530, opened 2026-06-03
RTECO-1287 jf poetry install --only main captures all dependency 

reshmifrog:
…ps in build info, ignoring group filters

---
Problem: jf poetry install all dependency ignoring group filters


utils/buildinfo/buildinfo.go
```diff
@@ -799,3 +824,66 @@ func extractRepoNameFromPypiURL(urlStr string) string {
...
+}
+
+// poetryInstalledPackages runs `poetry run pip list --format=json` inside
+// workingDir and returns the installed set as normalised name → version.
+// Normalisation matches PEP 503 (lowercase, runs of [-_.] collapsed to "-"),
+// the same normalisation applied to lockfile names by PoetryFlexPack.
+// Returns nil on any error so the caller falls back to lock-file-driven
+// resolution (no regression).
+func poetryInstalledPackages(workingDir string) map[string]string {
+	cmd := exec.Command("poetry", "run", "pip", "list", "--format=json")
```
fluxxBot: please try to use `poetry show` with all the flags from `poetry install` command being flushed into it

fluxxBot (changes requested):

utils/buildinfo/buildinfo.go
reshmifrog: 1. poetry show doesn't accept all poetry install flags. The reproduction command from the bug    report uses --no-root, and `$ poetry show --only main --no-root --format=json`
The option "--no-root" does not exist
Other install flags poetry show rejects: `--no-directory, --sync, --dry-run, --all-extras, --all-groups, --only-root, --compile, -E.` 

2. poetry show reads the lockfile, not the venv. Test on a venv with nothing installed:
   `$ poetry show --only main --format=json`
   [{"name": "numpy", "installed_status": "not-installed", "version": "1.26.1", ...}]
  It returns numpy even though numpy isn't installed. installed_status: "not-installed" is the giveaway  poetry           show is filtering the lockfile by group, not listing the venv. So it wouldn't reflect --no-root, partial-install recovery, or any non-flag-driven divergence.

fluxxBot (approved):

PR #3535, opened 2026-06-04
JGC-499 - Replace 'safe to test' label with single build-gate environment approval

ehl-jf:
## Summary

Replaces the per-workflow **`safe to test`** label gate with a single **`build-gate` GitHub Environment deployment approval**. A maintainer approves once per fork/PR run, and that single approval releases **frogbot** and **all integration-test suites**. Trusted `push` to `master` and manual `workflow_dispatch` runs skip the approval entirely.

## What changed

- **New `build-gate.yml` orchestrator** (`name: Build Gate`)
  - Triggers: `pull_request_target [opened, synchronize, reopened]`, `push` to `master`, `workflow_dispatch`.
  - A single `gate` job carries `environment: ${{ github.event_name == 'pull_request_target' && 'build-gate' || '' }}` — only fork/PR runs hit the environment and wait for approval; `push`/`workflow_dispatch` resolve to an empty environment (no gate).
  - 26 suite jobs (`frogbot` + `go` + 24 others), each `needs: gate` + `uses: ./.github/workflows/<suite>.yml` + `secrets: inherit`.
  - A `build-gate-success` aggregator job (`needs:` all suites, `if: always()`) that fails if any suite failed/was cancelled — see required-checks note below.
- **26 workflows converted to reusable** (`frogbot-scan-pull-request.yml` + 25 `*Tests.yml`)
  - `on:` → `workflow_call:` + `workflow_dispatch:` (dropped `push` and `pull_request_target: [labeled]`).
  - Removed the `safe to test` job conditions. `workflow_dispatch` inputs preserved (incl. poetry's `poetry_version`); `env:`/`permissions:` blocks intact.
  - `frogbot` no longer declares its own `environment: frogbot` — the gate now lives in the orchestrator.
  - `evidenceTests` and `ghostFrogTests` keep their existing `if: false` (independently disabled).
- **Deleted `removeLabel.yml`** — the label mechanism it served is gone.

## Security model

- Approval happens in `gate`, **before** any job checks out PR head code or touches secrets.
- Under `pull_request_target`, local `uses:` reusable workflows resolve from the **base** ref, so a PR cannot alter the gate or workflow logic — only the executed test/product code, which is exactly what the approval covers.

## Recovering a failed run

Use **"Re-run failed jobs"**: it re-runs the failed suite(s) **and** `build-gate-success` (reusing already-passed suites) while **not** re-running `gate`, so **no re-approval and no new commit** are needed — the single required check flips green on the same SHA. ("Re-run all jobs" re-runs `gate` and re-triggers approval.)

## :warning: Manual repo-settings changes required (not in this PR)

1. **Create the `build-gate` Environment** (Settings → Environments) with **Required reviewers** = the maintainer team.
2. **Verify frogbot secrets are repo/org-level**, not scoped to the old `frogbot` environment. frogbot no longer declares an environment and receives secrets via `secrets: inherit` (which passes repo/org secrets, **not** environment secrets). If `FROGBOT_URL`, `FROGBOT_ACCESS_TOKEN`, `JF_SMTP_*` were stored as `frogbot`-environment secrets, move them to repo/org level (or onto `build-gate`).
3. **Repoint branch-protection required status checks** to the single aggregator:
   - **`Build Gate / build-gate-success`**

   Remove the stale old per-workflow contexts. The aggregator is the only check that needs to be required — it transitively gates every suite, is matrix-independent (won't break when a suite's matrix changes), and stays pending until the approval gate is released.
4. *(Optional cleanup)* Delete the `safe to test` label and retire the now-unused `frogbot` environment.


barakharyati (approved):
.

.github/workflows/build-gate.yml
```diff
@@ -0,0 +1,183 @@
...
+  uv:
+    needs: gate
+    uses: ./.github/workflows/uvTests.yml
+    secrets: inherit
+
+  # Single, stable required status check. Point branch protection at
+  # "Build Gate / build-gate-success" instead of the matrix-expanded suite checks.
+  # Recover a failed suite with "Re-run failed jobs" (re-runs the suite + this job,
+  # not the approval gate) — no re-approval and no new commit needed.
+  build-gate-success:
```
barakharyati, line 161: Can we add gate also here?

.github/workflows/build-gate.yml
```diff
@@ -0,0 +1,183 @@
...
+    name: build-gate-success
+    if: always()
```
barakharyati, line 163: The if: always ()  cusing an issue here
build-gate-success doesn’t directly depend on gate, only on jobs after it. If gate is rejected, those jobs may be skipped, and the final check can still pass because it allows skipped jobs.

.github/workflows/build-gate.yml
```diff
@@ -0,0 +1,183 @@
...
+      - "master"
+  workflow_dispatch:
+
+# Ensures that only the latest commit is running for each PR at a time.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  pull-requests: write
```
barakharyati, line 21: It looks like there is a bug here (not security) 
I remember that the caller workflow cannot assign permissions it doesn't have  
In that case, the caller does not have  id-token: write, and it assigns it to a different workflow like .github/workflows/helmTests.yml by triggrering it 

https://github.com/jfrog/jfrog-cli/pull/3535/changes#diff-b4f30d948e9602b6e03aa68700eb3589067e5fabe411faa596842e3bcf0de387R22 row 22

barakharyati (commented):
Looks good 2 issues 1 security related and the second looks like a bug.

.github/workflows/build-gate.yml
ehl-jf: Done in `27645d29`. This also fixes the `if: always()` issue below.

.github/workflows/build-gate.yml
ehl-jf: Good catch. Fixed in `27645d29` by adding `gate` to `needs`: a rejected gate now fails the check, and an unapproved one keeps it pending. Kept `if: always()` so it still reports a definitive result when a suite fails.

.github/workflows/build-gate.yml
ehl-jf: Correct. Fixed in `27645d29` by granting `id-token: write` per-job on the four OIDC suites (`access`, `helm`, `huggingface`, `oidc`).

RemiBou (approved):

barakharyati:
LGTM we can merge it

PR #3560, opened 2026-06-19
Fix/latest release for hf breaks cli

reshmifrog:

---


BlasttS (approved):
Approved but I saw a ref to huggingface_hub install here : https://github.com/jfrog/jfrog-cli/blob/master/scripts/setup-test-deps.sh#L158
Even tho it's not blocking and an internal thing maybe it would be worth aligning the versin

PR #3561, opened 2026-06-19
RTECO-945: Wire jf apk commands and e2e tests

naveenku-jfrog:
## Summary

Wires the Alpine APK integration into the JFrog CLI frontend: registers `jf apk` subcommands, adds CLI flags, package aliasing (Ghost Frog), and a comprehensive e2e test suite. This is the user-facing entry point for the full Alpine APK feature.

Depends on:
- [build-info-go#393](https://github.com/jfrog/build-info-go/pull/393)
- [jfrog-cli-artifactory#496](https://github.com/jfrog/jfrog-cli-artifactory/pull/496)

---

### `buildtools/cli.go`
- **`ApkCmd()`** — dispatches `jf apk config`, `jf apk upload`, and all native `apk` subcommands (`add`, `upgrade`, etc.)
- Extracts JFrog-specific flags (`--repo`, `--server-id`, `--alpine-version`, `--user`, `--password`, `--build-name`, `--build-number`, `--module`, `--project`) before forwarding remaining args to native `apk`
- **Fail-fast on unknown `--server-id`**: if `--server-id` is explicitly provided but not found in the config, the command errors immediately — consistent with `jf npm`, `jf pip`, `jf mvn`, etc.
- Falls back to the default server silently when `--server-id` is omitted (allows `jf apk add` to work without Artifactory for basic usage)

### `utils/cliutils/commandsflags.go`
- New `Apk` flag group: `--repo`, `--server-id`, `--alpine-version`, `--user`, `--password`, `--build-name`, `--build-number`, `--module`, `--project`

### `docs/buildtools/apkcommand/help.go`
- `GetDescription()`, `GetAIDescription()`, `Usage`, `GetArguments()` for `jf apk` help text and AI help coverage

### `packagealias/packagealias.go`
- Registers `apk` as a Ghost Frog package alias so `jf apk add curl` works identically to `jfrog apk add curl` — consistent with `npm`, `pip`, `mvn`, etc.

### `apk_test.go`
Full end-to-end test suite (35 test cases) covering:

| Category | Tests |
|---|---|
| Basic build-info | `NoBuildFlags`, `WithBuildFlags`, `BuildNameFromEnvVars` |
| Dependency collection | `DepIDFormat`, `DepChecksums`, `DepRequestedBy`, `TransitiveDeps` |
| Scopes | `ScopeProduction`, `ScopeTransitive` |
| Repo / server config | `MissingRepo`, `InvalidRepo`, `UnknownServerID`, `ArtifactoryUnreachable`, `VirtualRepoAggregates` |
| Upload | `Upload`, `UploadWithBuildInfo`, `CIVCSPropertiesStamped` |
| Config | `Config`, `ConfigMissingServerID` |
| Env vars | `EnvVarsFiltered`, `EnvVarsIncluded` |
| Multi-package | `MultiPackageInstall`, `MultiPackageBuildInfo` |
| Idempotency | `IdempotentAdd` |
| Error handling | `ApkBinaryMissing` |

### `.github/workflows/apkTests.yml`
CI workflow that runs all `TestApk*` tests against a matrix of Alpine versions (`v3.18`, `v3.19`, `v3.20`, `v3.21`):
- Compiles the test binary on `ubuntu-24.04` (where Go 1.26 is available)
- Runs the pre-built static binary inside each `alpine:<version>` Docker container to get a real `apk` binary
- Spins up a local Artifactory instance via `install-local-artifactory`
- Triggered via `workflow_call` and `workflow_dispatch`

### `scripts/test-apk-local.sh`
Helper script for running Alpine APK e2e tests locally on macOS via Docker (mounts the repo and `~/.jfrog` config into a `golang:alpine` container).

---

## Design Notes

**`--repo` is optional**: Consistent with `jf npm install` and `jf pip install`, omitting `--repo` does not block build-info collection. Checksums are always available from the APK database (`C:` field). AQL enrichment (for SHA-256/MD5) is skipped when no repo is specified.

**Test infrastructure**: Tests require the `apk` binary (Alpine Linux). The workflow compiles a static Linux binary on the host and runs it inside Alpine Docker containers, avoiding Go version constraints in Alpine package repos.

**AI help coverage**: `jf apk` is wired through `corecommon.ResolveDescription` to satisfy the `TestAIHelpCoverageGenerated` check.


buildtools/cli.go
```diff
@@ -2118,6 +2132,119 @@ func NixCmd(c *cli.Context) error {
...
+		// User explicitly named a server — fail fast if it doesn't exist.
+		serverDetails, err = coreConfig.GetSpecificConfig(serverID, false, false)
+		if err != nil {
+			return errorutils.CheckErrorf("server ID %q not found in configuration. "+
+				"Run 'jf c add' to add it, or omit --server-id to use the default server.", serverID)
+		}
+	} else {
+		serverDetails, err = coreConfig.GetDefaultServerConf()
+		if err != nil {
+			log.Warn("No JFrog server configured — skipping credential injection. Run: jf c add")
```
agrasth: `coreConfig.GetDefaultServerConf()` returns `(nil, nil)` when no server is configured at all (the common case for a user who hasn't run `jf c add`), so this `err != nil` branch won't actually fire there and `serverDetails` stays nil silently. Worth double-checking what this warning is meant to cover.

buildtools/cli.go
```diff
@@ -2118,6 +2132,119 @@ func NixCmd(c *cli.Context) error {
...
+	case "upload":
+		return apkUploadSubCmd(c, remainingArgs, serverDetails, repoKey, alpineVersion, username, password)
+	default:
+		filteredArgs, buildConfiguration, err := build.ExtractBuildDetailsFromArgs(remainingArgs)
+		if err != nil {
+			return err
+		}
+		cmd := alpinecommand.NewApkCommand(subcmd).
+			SetArgs(filteredArgs).
+			SetServerDetails(serverDetails).
```
agrasth: `SetServerDetails(serverDetails)` is called unconditionally here, even when `serverDetails` is nil (see the fallback branch above). `NixCmd` just above (line 2128-2130) guards this with `if err == nil && serverDetails != nil`. Should `ApkCmd` follow the same guard for consistency / to avoid passing a nil server details struct downstream?

.github/workflows/apkTests.yml
```diff
@@ -80,16 +84,64 @@ jobs:
...
+              -w /workspace \
+              "alpine:${ALPINE_VER#v}" \
+              ./apk.test \
+                -test.v -test.timeout 0 \
+                -test.run "$run_regex" \
+                -test.alpine \
+                $creds 2>&1 | tee /tmp/apk-out.log
+
+            # Collect the top-level names of tests that failed this attempt.
+            fails=$(grep -Eo '^--- FAIL: [A-Za-z0-9_]+' /tmp/apk-out.log | awk '{print $3}' | sort -u | tr '\n' '|' | sed 's/|$//')
```
agrasth, line 138: This only checks for literal `--- FAIL:` lines in the captured output — it doesn't check `docker run`'s own exit code. If the test binary crashes/panics or the container gets OOM-killed before printing any `--- FAIL:` line, `$fails` stays empty and the step reports "All Alpine tests passed" on that attempt. Might be worth capturing `${PIPESTATUS[0]}` (or splitting the pipe) and treating a non-zero docker exit as a failure too.

apk_test.go
```diff
@@ -0,0 +1,1905 @@
...
+	jfrogCli := coretests.NewJfrogCli(execMain, "jfrog", "")
+	err := jfrogCli.Exec("apk", "add", "curl",
+		"--repo="+tests.AlpineVirtualRepo,
+		"--server-id=nonexistent-server-id-xyz")
+	assert.Error(t, err, "jf apk add with an unknown --server-id should return an error")
+}
+
+// TestApkAdd_ArtifactoryUnreachable verifies that when Artifactory is unreachable
+// jf apk add returns a clear error and does not silently fall back to the public CDN.
+func TestApkAdd_ArtifactoryUnreachable(t *testing.T) {
```
agrasth, line 1800: This test body is identical to `TestApkAdd_UnknownServerID` above (same `--server-id=nonexistent-server-id-xyz` args) — it doesn't actually simulate an unreachable Artifactory server, it just re-tests the unknown-server-id path. Should this point at a real unreachable host/port instead (e.g. an invalid URL in a throwaway server config) to cover the case the name describes?

apk_test.go
```diff
@@ -0,0 +1,1905 @@
...
+		for _, dep := range publishedBuildInfo.BuildInfo.Modules[0].Dependencies {
+			if dep.Sha1 != "" || dep.Sha256 != "" {
+				depsWithChecksums++
+			}
+		}
+		t.Logf("Dependencies with at least one checksum: %d/%d",
+			depsWithChecksums, len(publishedBuildInfo.BuildInfo.Modules[0].Dependencies))
+		// On a machine with a populated APK cache, checksums should be present.
+		// We assert at least the package itself has one.
+		assert.Greater(t, depsWithChecksums, 0,
```
agrasth: This only asserts `dep.Sha1 != "" || dep.Sha256 != ""` — `Md5` is never checked anywhere in this test. Per the checksum requirement (sha256/sha1/md5 all expected), can we add an assertion that at least one dep has a non-empty `Md5` too?

apk_test.go
```diff
@@ -0,0 +1,1905 @@
...
+			for _, scope := range dep.Scopes {
+				if scope == "prod" {
+					hasProd = true
+				}
+				if scope == "transitive" {
+					hasTransitive = true
+				}
+			}
+		}
+		assert.True(t, hasProd, "at least one dependency should have scope 'prod'")
```
agrasth, line 467: `hasTransitive` is computed but only logged at line 468-469, never asserted. Given the comment above says curl reliably brings in transitive deps (libcurl, musl, etc.), should this be `assert.True(t, hasTransitive, ...)` like the `hasProd` check right above it, rather than just a log line?

apk_test.go
```diff
@@ -0,0 +1,1905 @@
...
+	assert.True(t, found)
+
+	if found && len(publishedBuildInfo.BuildInfo.Modules) > 0 {
+		deps := publishedBuildInfo.BuildInfo.Modules[0].Dependencies
+		if len(deps) <= 1 {
+			t.Skip("only one dep installed; transitive RequestedBy cannot be validated")
+		}
+		hasTransitiveRequestedBy := false
+		for _, dep := range deps {
+			if len(dep.RequestedBy) > 0 && len(dep.RequestedBy[0]) > 0 {
```
agrasth: This only checks that a `RequestedBy` chain is non-empty, not that its content actually identifies the correct parent (e.g. that libcurl's chain contains `curl:<version>`). Could tighten this to assert on the actual parent ID rather than just presence.

docs/buildtools/apkcommand/help.go
```diff
@@ -0,0 +1,75 @@
...
+
+var Usage = []string{
+	"apk upload <file.apk> --repo <repo-key> --alpine-version <vX.Y> [--branch <name>] [--arch <arch>] [jf-flags]",
+	"apk add <packages...> [jf-flags]",
+	"apk upgrade [packages...] [jf-flags]",
+	"apk update [jf-flags]",
+	"apk fetch <packages...> [jf-flags]",
+	"apk search <pattern> [jf-flags]",
+	"apk del <packages...>",
+	"apk info [packages...]",
```
agrasth: This `Usage` list (and the subcommand list in `GetArguments()` below) only documents upload/add/upgrade/update/fetch/search/del/info. `apk_test.go`'s `TestApkPassthroughCommands` explicitly exercises `fix`, `audit`, `version`, and `stats` as supported passthrough subcommands — should those be documented here too so `jf apk --help` reflects everything that's actually supported?

utils/cliutils/commandsflags.go
```diff
@@ -2243,6 +2256,9 @@ var commandFlags = map[string][]string{
 	Nix: {
 		BuildName, BuildNumber, module, Project, serverId,
 	},
+	Apk: {
+		serverId, repo, apkAlpineVersion, branch, apkArch, BuildName, BuildNumber, module, Project, user, password,
```
agrasth: This reuses the shared `branch` flag, whose `Usage` text (line 1855) says `"[Mandatory] Branch name to filter."`. For `jf apk upload`, `--branch` is actually optional and defaults to `"main"` (per `apkUploadSubCmd` in buildtools/cli.go). Since this flag's help text is shared across commands, `jf apk upload --help` will show a misleading "[Mandatory]" for an optional flag — might need a per-command override or a tweaked shared description.

.github/workflows/apkTests.yml
naveenku-jfrog: Good catch — fixed. The step now captures the container exit code via `docker_exit=${PIPESTATUS[0]}` and only reports success when both there are no `--- FAIL:` lines **and** `docker_exit -eq 0`. A non-zero exit with no test-level failures (crash/panic/OOM) is now explicitly treated as a failure for that attempt.

apk_test.go
naveenku-jfrog: Fixed — this test now configures a server that points at a genuinely unreachable address (`--artifactory-url=http://127.0.0.1:1/artifactory`) rather than reusing the unknown-server-id path, so it exercises the connection-failure case the name describes.

apk_test.go
naveenku-jfrog: Fixed — `hasTransitive` is now asserted with `assert.True(t, hasTransitive, ...)` (guarded by `len(deps) > 1`) instead of only being logged, mirroring the `hasProd` assertion above it.

reshmifrog (approved):

PR #3592, opened 2026-07-08
RTECO-945 - Fix Go version mismatch in Alpine APK CI

naveenku-jfrog:
Alpine package repos only ship Go <=1.23 (even on Alpine 3.21), but go.mod requires go 1.26.3, causing 'invalid go version' parse errors.

Fix: do not install Go via 'apk add go'. Instead:
- Record the host GOROOT after install-go-with-cache (new step)
- Mount that directory into the Alpine chroot via jirutka/setup-alpine volumes input so the same Go 1.26 binary is available inside Alpine
- Export GOROOT/GOPATH/GOCACHE into the alpine.sh test step and prepend GOROOT/bin to PATH so 'go' resolves to the host toolchain

---

### DIFF
--- a/.github/workflows/apkTests.yml
+++ b/.github/workflows/apkTests.yml
@@ -73,9 +73,6 @@ jobs:
       - name: Setup Go with cache
         uses: jfrog/.github/actions/install-go-with-cache@main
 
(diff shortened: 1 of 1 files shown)


attiasas (approved):
LGTM

PR #3596, opened 2026-07-09
RTECO-1564 - Add apt command support for JFrog Artifactory

fluxxBot:

---
- Introduced `apt` command to run apt-get commands against JFrog Artifactory Debian repositories.
- Implemented `aptSetupCmd` for persistent authentication setup.
- Added necessary flags and help documentation for the `apt` command.
- Created test data for local, remote, and virtual APT repositories.
- Updated integration tests to include APT command tests.
- Enhanced CLI utility functions to support APT-specific flags and commands.


naveenku-jfrog (approved):

buildtools/cli.go
```diff
@@ -2150,6 +2172,169 @@ func NixCmd(c *cli.Context) error {
...
+
+	component := c.String("component")
+	importKey := c.Bool("import-key")
+	trusted := c.Bool("trusted")
+
+	// Interactive fallback: prompt for component and GPG mode when dist was not
+	// supplied as a flag (i.e., the user is in an interactive session and hasn't
+	// pre-scripted all flags).
+	if interactive {
+		if component == "" {
```
agrasth, line 2315: This `if component == ""` branch is unreachable in practice: the `component` flag is defined with `Value: "main"` (see `aptComponent` in `commandsflags.go`), so `c.String("component")` returns `"main"` by default even when the user never passed `--component` — `component` is never `""` here. 

The interactive "Component ... leave empty for 'main'" prompt described in the comment above can never actually show. Either drop the default `Value` on the flag (so an unset flag reads as `""` and the prompt fires as intended) or remove this dead prompt.

agrasth (approved):

PR #3615, opened 2026-07-23
RTECO-1647 - Add support for Maven Wrapper (mvnw) command

fluxxBot:

---
- Introduced a new command "mvnw" in cli.go to run Maven builds using the project's Maven Wrapper.
- Created help documentation for the mvnw command in help.go.
- Updated go.mod to include a dependency on jfrog-cli-artifactory.
- Added tests for the new mvnw command in maven_test.go to validate its behavior in native mode.
- Included necessary Maven Wrapper files (MavenWrapperDownloader.java, maven-wrapper.jar, maven-wrapper.properties, mvnw, mvnw.cmd) in the test data for proper execution.
- Created a sample pom.xml for testing Maven projects with the wrapper.


buildtools/cli.go
```diff
@@ -664,7 +680,19 @@ func captureUserFlagsForMetrics(c *cli.Context, skipFlagParsing bool) {
...
+	return runMvn(c, false)
+}
+
+// MvnwCmd runs "jf mvnw". In native (FlexPack) mode it requires a Maven Wrapper
+// (mvnw/mvnw.cmd) to be present and fails otherwise; in legacy (config-file) mode
+// it behaves exactly like MvnCmd.
+func MvnwCmd(c *cli.Context) (err error) {
+	return runMvn(c, true)
+}
+
```
naveenku-jfrog, line 694: In legacy (config-file) mode preferWrapper is silently dropped and this runs PATH mvn even though the user typed mvnw. Can we emit a log.Debug/log.Info so the divergence is discoverable rather than silent?

naveenku-jfrog (approved):

buildtools/cli.go
```diff
@@ -688,7 +716,7 @@ func MvnCmd(c *cli.Context) (err error) {
 		if err != nil {
 			return err
 		}
-		mvnCmd := mvn.NewMvnCommand().SetConfigPath("").SetGoals(filteredMavenArgs).SetConfiguration(buildConfiguration).SetServerDetails(serverDetails)
+		mvnCmd := mvn.NewMvnCommand().SetConfigPath("").SetGoals(filteredMavenArgs).SetConfiguration(buildConfiguration).SetServerDetails(serverDetails).SetPreferWrapper(preferWrapper)
```
agrasth, line 719: `preferWrapper` only reaches `MvnCommand` through this native-mode branch (`ShouldRunNative(configFilePath) && !configExists`, a few lines above). 

When a legacy Maven config file already exists for the project, `jf mvnw` falls through to the legacy path lower in this function, where `preferWrapper` is never consulted — so `jf mvnw` silently behaves exactly like `jf mvn` in that case. That's documented in `docs/buildtools/mvnw/help.go`'s AI description, but nothing is logged at the point of the fallback. Might be worth a `log.Debug` noting that wrapper preference is being ignored because a legacy config was found, for troubleshooting a case where a user expects `mvnw` semantics but the project still has a `.jfrog/projects/maven.yaml`.

agrasth (approved):

PR #3618, opened 2026-07-27
RTECO-1651 - Update agent plugins tests

udaykb2:

---
## Summary
Expands and hardens `jf agent plugins` e2e coverage in `agent_plugins_test.go` (85 → 99 tests), and fixes the agent-plugins CI workflow so tests run reliably on Linux/Windows.

### What the suite covers
| Area | Focus |
|------|--------|
| **Publish** | Basic publish, version override, collisions (CI), invalid semver/slug, wrong/missing repo, prebuilt zip, build-info/module/env vars, signing key, checksum/`agentpackages.digest`, multi-harness marketplace indexing |
| **Install** | Latest / specific version, marketplace (with `--harness`), global/path scopes, multi-harness (`claude`/`codex`/`cursor`), `plugin-info.json`, evidence gate, JSON format, flag mutual exclusion, agent-config override, custom `my-agent` lifecycle |
| **Update** | Per-slug, `--all`, dry-run, force, non-interactive, JSON format, project-scope rejection, flag matrix |
| **List** | Local/remote, latest-only, empty local, `--check-updates` statuses, limit (harness/repo), flag validation |
| **Search** | Exact/substring, latest-only, no matches, empty/blank query, repo from env |
| **Delete** | Delete, dry-run (incl. multi-version / not found), version-specific, missing args, repo from env |
| **Repo / server / proxy** | Env vs `--repo`, no-repo error, `--server-id`, unreachable Artifactory, proxy / no-proxy, insecure TLS |
| **Round-trips / CI** | Publish→install→update→delete flows, CI pipeline path |

### Notable test improvements
- Stronger error assertions (`assertErrorContainsAll` instead of loose OR matches)
- Checksum / Artifactory property verification
- Marketplace install prefers `--harness` and retries until marketplace indexes
- Custom agent install/update/list/delete lifecycle e2e
- Shared harness matrix for claude / codex / cursor (and multi-harness)

### CI / workflow
- `agentPluginsTests.yml`: use `-test.agentPlugins=true` (Windows-safe), force `bash` shell
- Stop pinning Artifactory VERSION (avoids pruned release breakage)
- Gosec: suppress G703 on test stub binary write under `t.TempDir`


agent_plugins_test.go
```diff
@@ -140,23 +156,262 @@ func pluginArtifactPath(repo, slug, version string) string {
...
+	data, err := os.ReadFile(claudeBin) // #nosec G304 -- path from t.TempDir
 	require.NoError(t, err)
-	require.NoError(t, f.Close())
-	t.Cleanup(func() { _ = os.Remove(f.Name()) })
-	fileName := harness + "-marketplace.json"
-	require.NoError(t, artifactoryCli.Exec("u", f.Name(),
-		tests.AgentPluginsLocalRepo+"/"+fileName,
-		"--flat=true",
-	), "uploading %s to test repo must succeed", fileName)
+	require.NoError(t, os.WriteFile(codexBin, data, 0755)) // #nosec G306,G703 -- test stub binary path is under t.TempDir
```
naveenku-jfrog: G703 isn't a valid gosec rule ID (gosec is G1xx–G6xx); unknown IDs are silently ignored so it's harmless, but it looks like a typo. I think you only need G306 here — drop G703.

agent_plugins_test.go
```diff
@@ -43,12 +48,15 @@ func CleanAgentPluginsTests() {
...
-	t.Skip("Agent plugins e2e tests are disabled")
+	if !*tests.TestAgentPlugins {
+		t.Skip("Skipping Agent Plugins test. To run Agent Plugins test add the '-test.agentPlugins=true' option.")
+	}
 	createJfrogHomeConfig(t, false)
 	require.True(t, isRepoExist(tests.AgentPluginsLocalRepo), "agent plugins local repo does not exist: "+tests.AgentPluginsLocalRepo)
 	// The test Artifactory instance has no evidence/One-Model service configured.
 	// Disable the quiet-failure evidence gate so install commands don't block on 403.
 	t.Setenv("JFROG_AGENT_PLUGINS_DISABLE_QUIET_FAILURE", "true")
+	stubNativeAgentCLIs(t)
```
naveenku-jfrog, line 60: stubNativeAgentCLIs runs a full go build on every test init, so the stub is recompiled ~99 times across the suite (likely a big chunk of the ~11m Windows run). Consider building the stub once (sync.Once or TestMain) and reusing the binary across tests

agent_plugins_test.go
```diff
@@ -2240,22 +3008,25 @@ func TestAgentPluginsRepoFlagOverridesEnvVar(t *testing.T) {
...
 func TestAgentPluginsNoRepoConfigured(t *testing.T) {
 	initAgentPluginsTest(t)
 	defer cleanAgentPluginsTest()
 
-	// Ensure env var is not set so there is no fallback.
 	t.Setenv("JFROG_AGENT_PLUGINS_REPO", "")
 
+	// Temporarily remove the suite's agentplugins repo so auto-discovery finds nothing.
+	require.True(t, isRepoExist(tests.AgentPluginsLocalRepo))
+	execDeleteRepo(tests.AgentPluginsLocalRepo)
```
naveenku-jfrog, line 3081: This deletes the shared suite repo and recreates it via t.Cleanup. Safe only because the suite is sequential — if anyone adds t.Parallel() later, or the process hard-exits before cleanup, every subsequent test loses the repo. Worth a comment stating the sequential-only assumption.

agent_plugins_test.go
```diff
@@ -911,7 +1207,8 @@ func TestAgentPluginsInstallAgentConfigOverride(t *testing.T) {
 		}
 	}`)
 
-	// Verify globalDir override.
+	// Custom agents are not indexed into <harness>-marketplace.json, so --version is
```
naveenku-jfrog, line 1248: The old "no --project-dir and no --global → defaults to cwd" case was dropped rather than updated to the new default-to---global behavior, and the custom-agent lifecycle test doesn't cover that implicit/default-scope path. Intentional? If the default scope changed, it'd be good to assert the new default explicitly.

naveenku-jfrog (approved):
Please address some minor comments.

agent_plugins_test.go
udaykb2: Done

agent_plugins_test.go
udaykb2: done, added

agent_plugins_test.go
udaykb2: DONE, now, go build runs once via sync.Once (nativeAgentStubBinary).
Each test only copies that binary into its temp PATH dir (stubNativeAgentCLIs).

PR #3632, opened 2026-07-31
Surface the invoking AI agent on the CLI's User-Agent

noamshemesh:
## What

`jfrog-cli-core` already detects which AI harness invoked the CLI, from a table of harness
environment variables — Claude Code, Gemini, Goose, Cursor, Copilot, Kilocode, Roo Code, Codex,
plus a generic `AGENT` fallback collapsed to `unknown` (`common/commands/execution_context.go`).

That answer never reached the wire. Its five call sites are telemetry labels, interactive-prompt
suppression, AI-formatted help, the survey link, and the Cursor trace ID — none of them touch the
outbound request. So an agent running `jf rt download` and a human running it are byte identical to
anything in the request path.

This appends the detected agent to the CLI's User-Agent as an RFC 9110 comment:

```
User-Agent: jfrog-cli-go/2.117.0 (agent:claude)
```

## Why a comment rather than a new header or a rewritten token

- **The product token stays first**, so anything parsing only the leading token is unaffected.
- **A non-agent invocation is byte identical to today** — no behaviour change for human use.
- **A dedicated header would need `jfrog-client-go`.** The only global per-request header hook there
  is `addUberTraceIdHeaderIfSet`, so a new header means a second repo and a release the CLI then
  bumps to. That remains the durable form; this lands the same signal now with no new plumbing.
- `JFROG_CLI_USER_AGENT` overrides are preserved — the marker is appended to whatever the override
  resolves to, never replacing it.

Worth noting: `execution_context.go` already lists "User-Agent enrichment" among its call sites in
the memoization comment, but no such call site existed. This looks like intent that never landed.

## Safety

Injection-safe by construction: `DetectExecutionContext` returns a name from a fixed table or the
literal `unknown`, so a raw environment value is never propagated onto the wire. A test asserts no
CR/LF can appear.

**The value is attribution metadata, not a credential.** Harness environment variables are
client-set and trivially forged or unset. Consumers must treat it as a routing/census hint only —
it must never gate authorization.

## Tests

15 unit tests in `utils/cliutils`:

- one case per row of the `agentEnvDetectors` table, plus the generic `AGENT` → `unknown` fallback
- no agent detected → User-Agent byte identical to today
- `JFROG_CLI_USER_AGENT` override preserved with the marker appended
- no-version case (`GetCliUserAgent` omits the slash)
- well-formedness: product token first, no CR/LF

Tests clear the agent env vars and reset the memoized execution context, so they are deterministic
even when the suite itself runs inside Claude Code or Cursor.

`go build ./...` and `go vet` clean.

## Downstream

JFrog Agent Gateway consumes this marker to attribute diverted requests to a specific harness
rather than a generic "agent", verified end-to-end against a live JPD. That side is already open as
a separate PR and is inert until this ships.

### DIFF
--- a/utils/cliutils/utils_test.go
+++ b/utils/cliutils/utils_test.go
@@ -494,3 +494,97 @@ func TestTransferFilesTimestampFilterFlags(t *testing.T) {
 	assert.Contains(t, usageByName[CreatedAfter], "YYYY-MM-DDTHH:mm:ss.sssZ")
 	assert.Contains(t, usageByName[DownloadedAfter], "YYYY-MM-DDTHH:mm:ss.sssZ")
 }
+
+// --- AGW-86: User-Agent enrichment with the detected AI agent ---
+
+// withCliUserAgent pins the CLI user-agent name/version for one test. The real values are
+// set once in init() from JFROG_CLI_USER_AGENT, so tests drive the setters directly rather
+// than trying to re-run init.
+func withCliUserAgent(t *testing.T, name, version string) {
+	t.Helper()
+	prevName, prevVersion := coreutils.GetCliUserAgentName(), coreutils.GetCliUserAgentVersion()
+	coreutils.SetCliUserAgentName(name)
+	coreutils.SetCliUserAgentVersion(version)
+	t.Cleanup(func() {
+		coreutils.SetCliUserAgentName(prevName)
+		coreutils.SetCliUserAgentVersion(prevVersion)
+	})
+}
+
+func TestGetCliUserAgentWithAgentNoAgentDetected(t *testing.T) {
+	clearAgentEnvVarsForTest(t)
+	withCliUserAgent(t, "jfrog-cli-go", "2.117.0")
+	corecommands.ResetExecutionContextForTest()
+
+	assert.Equal(t, "jfrog-cli-go/2.117.0", GetCliUserAgentWithAgent(),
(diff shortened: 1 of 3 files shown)


RemiBou (approved):

utils/cliutils/utils_test.go
```diff
@@ -494,3 +494,97 @@ func TestTransferFilesTimestampFilterFlags(t *testing.T) {
...
+
+func TestGetCliUserAgentWithAgentPreservesCustomUserAgent(t *testing.T) {
+	// JFROG_CLI_USER_AGENT lets an operator replace the product token entirely. The agent
+	// marker must be appended to whatever that resolves to, never replace it.
+	clearAgentEnvVarsForTest(t)
+	withCliUserAgent(t, "my-wrapper", "9.9.9")
+	t.Setenv("CLAUDECODE", "true")
+	corecommands.ResetExecutionContextForTest()
+
+	assert.Equal(t, "my-wrapper/9.9.9 (agent:claude)", GetCliUserAgentWithAgent())
```
RemiBou: isnt "agent/claude" more conventional ?

noamshemesh: maybe also ai-agent/claude ? wdyt?

RemiBou: yes better

RemiBou (approved):

noamshemesh:

PR #3634, opened 2026-08-01
RTECO-1651 - Add agent skills e2e tests

udaykb2:
Cover jf agent skills publish/install/update/delete/list/search flows with CI wiring and dependency pins for the skills Artifactory API.

---
## Summary
Adds `jf agent skills` e2e coverage in `agent_skills_test.go` (~20 top-level suites with harness/flag subtests), wires a Skills local repo + `-test.agentSkills`, and hooks an `agent-skills` job into the build gate so Linux/Windows CI runs the suite.

## What the suite covers

| Area | Focus |
|---|---|
| Publish | Happy path + checksums, frontmatter version, `--version` override, quiet/CI collision, prebuilt `zip/{slug}_{version}.zip` layout, validation (missing `SKILL.md` / name / frontmatter, invalid slug/semver, missing path), build-info module + build props |
| Install | Project harness matrix (cursor, claude-code, github-copilot, windsurf, codex, cross-agent, multi-harness), global cursor, latest via `--path`, path vs harness mutual exclusion, unknown harness, missing skill/version, evidence quiet-failure gate (`JFROG_SKILLS_DISABLE_QUIET_FAILURE`), `.jfrog/skill-info.json` |
| Update | Dry-run, update to latest, `--force` + JSON summary, published-but-not-installed (summary detail), not-in-repo |
| Delete | Dry-run, delete one version (storage absence via `FileInfo`), missing `--version` |
| List | `--repo` JSON (`Repo: ` source prefix), local harness list after install, empty local (no skills dir), flag matrix (neither mode / mutual exclusion / comma harness / `--limit=0`), `--limit` on repo list, `--check-updates` behind → current |
| Search | Match, `--prop` match, no matches, empty/blank query |
| Repo resolution | `JFROG_SKILLS_REPO` env, nonexistent repo (`upload failed`), multi-repo quiet/CI fail |
| Custom agent | Install/list/update lifecycle under `skills-agents` in `agent-config.json` |
| Round-trips / CI | Publish→install byte round-trip of `SKILL.md`; quiet CI publish(+bp)→install→list→update |
| Flags | Unknown `--format` falls back to table; unknown flags per subcommand; missing install slug |

## Notable test improvements
- Skills API indexing lag handled with shared backoff (`waitForSkillIndexed`, list/search retries) — storage upload alone is not enough for install/list/search/update/delete-dry-run
- Stronger error assertions via `assertErrorContainsAll` (no loose OR matches)
- Artifact presence/absence uses storage `FileInfo` (not `GetItemProps`, which maps “no properties” 404 → nil success)
- Evidence gate exercised with/without `JFROG_SKILLS_DISABLE_QUIET_FAILURE` (test Artifactory has no One-Model)
- Windows-safe path assertion: list JSON always uses forward slashes (`.cursor/skills/<slug>`)
- Shared harness matrix mirrors skills install layout so e2e fails if agent path maps drift

## CI / workflow
- New `agentSkillsTests.yml`: matrix ubuntu/windows/macos (macOS skipped — JGC-413), `go test … --test.agentSkills`
- `build-gate.yml`: `agent-skills` callable job
- Pins `jfrog-cli-artifactory` @ `5dfbbf6` and compatible `jfrog-cli-core` for Skills API support
- Fixture: `testdata/skills_local_repository_config.json` (`packageType: skills`) + `testdata/agent_skills/test-skill`


agent_skills_test.go
```diff
@@ -0,0 +1,2751 @@
...
+		t.Setenv("CI", "true")
+		t.Setenv("JFROG_SKILLS_DISABLE_QUIET_FAILURE", "")
+		err := runAgentSkillsCmd(t,
+			"install", slug,
+			"--repo="+tests.AgentSkillsLocalRepo,
+			"--path="+t.TempDir(),
+			"--quiet",
+		)
+		if err == nil {
+			t.Skip("evidence quiet-failure gate not enforced on this Artifactory instance")
```
naveenku-jfrog: This t.Skip fires whenever the gate doesn't trigger, and the suite header notes the shared test Artifactory has no evidence/One-Model service — so on CI this negative-path assertion ("evidence verification failed") is almost always skipped and gives no real signal. Can you confirm the evidence-failure path is covered elsewhere (e.g. a unit test in jfrog-cli-artifactory)? If not, consider gating this behind an explicit env flag so it's clear when it's meant to actually run.

agent_skills_test.go
```diff
@@ -0,0 +1,2751 @@
...
+
+	t.Setenv("CI", "true")
+	t.Setenv("JFROG_SKILLS_DISABLE_QUIET_FAILURE", "")
+	err := runAgentSkillsCmd(t,
+		"install", slug,
+		"--repo="+tests.AgentSkillsLocalRepo,
+		"--path="+t.TempDir(),
+		"--quiet",
+	)
+	if err == nil {
```
naveenku-jfrog: Same skip pattern as the evidence-gate test — on the shared instance this branch is effectively a no-op. Worth a comment stating that's expected, and confirming the assertion runs somewhere.

agent_skills_test.go
```diff
@@ -0,0 +1,2751 @@
...
+		"downloaded zip sha256 must match Artifactory-stored checksum")
+}
+
+// TestAgentSkillsArtifactoryUnreachable covers plan #85.
+func TestAgentSkillsArtifactoryUnreachable(t *testing.T) {
+	initAgentSkillsTest(t)
+	defer cleanAgentSkillsTest()
+
+	const bogusServerID = "unreachable-skills-rt"
+	bogusServerURL := "https://nonexistent-artifactory-host-xyzzy.example.com/artifactory/"
```
naveenku-jfrog: Relying on *.example.com failing to resolve is nondeterministic — some CI resolvers/captive DNS will hang for the full connection timeout, and the assertion is only assert.Error. Suggest an unroutable TEST-NET address (https://192.0.2.1/artifactory/, RFC 5737) plus a short connection timeout so the failure is fast and deterministic.

agent_skills_test.go
```diff
@@ -0,0 +1,2751 @@
...
+	initAgentSkillsTest(t)
+	defer cleanAgentSkillsTest()
+
+	t.Setenv("JFROG_SKILLS_REPO", "")
+
+	// Temporarily remove the suite's shared Skills repo so auto-discovery finds nothing.
+	// Safe only while this package runs sequentially (no t.Parallel): a hard process exit
+	// before Cleanup would leave later tests without the suite repo.
+	require.True(t, isRepoExist(tests.AgentSkillsLocalRepo))
+	execDeleteRepo(tests.AgentSkillsLocalRepo)
```
naveenku-jfrog, line 1595: This deletes the suite-wide skills repo and recreates it in t.Cleanup. It's well-documented, but a hard process exit between delete and cleanup cascades into every later test. Since the whole suite depends on this one shared repo, can we accept this risk explicitly, or scope the "no repo" scenario to a throwaway repo instead of mutating the shared one?

agent_skills_test.go
```diff
@@ -0,0 +1,2751 @@
...
+	Detail string `json:"detail"`
+}
+
+type agentSkillsSummaryJSON struct {
+	Slug    string                     `json:"slug"`
+	Version string                     `json:"version"`
+	Results []agentSkillsSummaryResult `json:"results"`
+}
+
+func parseSkillsSearchRows(out string) ([]agentSkillsSearchRow, error) {
```
naveenku-jfrog, line 480: This rolls its own […] extraction while the rest of the file uses extractJSONPayload. Two heuristics for the same "pull JSON out of log-polluted stdout" problem — consider consolidating on extractJSONPayload for consistency.

agent_skills_test.go
```diff
@@ -0,0 +1,2751 @@
...
+	require.NoError(t, biutils.CopyDir(skillSrc, skillPath, true, nil))
+	writeSkillMD(t, skillPath, slug, version, "Integration test skill")
+	return skillPath
+}
+
+func writeSkillMD(t *testing.T, skillDir, slug, version, description string) {
+	t.Helper()
+	content := fmt.Sprintf("---\nname: %s\ndescription: %s\nversion: %s\n---\n\n# %s\n\nBody content for agent skills e2e.\n",
+		slug, description, version, slug)
+	require.NoError(t, os.WriteFile(filepath.Join(skillDir, "SKILL.md"), []byte(content), 0644)) // #nosec G306 -- test fixture
```
naveenku-jfrog: Nit: this writes SKILL.md at 0644 while most other fixtures use 0600. Harmless, just inconsistent — worth aligning.

.github/workflows/agentSkillsTests.yml
```diff
@@ -0,0 +1,58 @@
+name: Agent Skills Tests
```
naveenku-jfrog, line 1: --timeout 0 with ~40 tests, each publishing multiple versions and waiting on indexing (90s budget each), across 3 OSes with fail-fast: false — this could be a long/expensive job. Not blocking, but flagging the CI cost.

naveenku-jfrog (approved):
Some minor improvements comments.

.github/workflows/agentSkillsTests.yml
udaykb2: The 90-second indexing budget (`skillsIndexBudget`) is a worst-case cap. The retry helper (`retryWithBackoffSkills`) uses exponential backoff starting at 2 seconds and caps at 10-second sleeps. In normal execution, the API usually sees the newly published skill versions in 2–4 seconds, meaning most of the ~40 tests resolve almost instantly without consuming the budget

agent_skills_test.go
udaykb2: Done, We eliminated the custom parsing code and consolidated both methods onto a single parser helper (`findJSONPayload`) that uses the robust newline-bound extraction logic. 

We updated `parseSkillsSearchRows`, our test wrappers, and retry helpers to use it, completely removing the duplicate heuristics.

udaykb2: Done! We eliminated the custom parsing code and consolidated both methods onto a single parser helper () that uses the robust newline-bound extraction logic. 

We updated , our test wrappers, and retry helpers to use it, completely removing the duplicate heuristics.

.github/workflows/agentSkillsTests.yml
udaykb2: Thanks for flagging this! 

To keep this suite completely consistent with existing package test conventions and avoid deviation, we decided to keep  (like , , , etc.) and let standard global runner timeouts govern the execution. 

Also, since we use exponential backoff polling rather than static sleeps, the indexing usually completes and exits in 2-4 seconds on healthy runs, meaning the 90s budget is rarely spent.

agent_skills_test.go
udaykb2: We have to delete the shared repo because auto-discovery scans every repository of the skills package type, so a separate throwaway repository wouldn't isolate the test.

To eliminate the cascading failure risk, we made the suite **self-healing**: `initAgentSkillsTest` (which runs at the start of every test) now checks if the shared repository is missing and automatically recreates it if needed. This prevents any crashed runs from breaking subsequent test executions.

PR #3640, opened 2026-08-04
Added Post Failure Run support for pipenv

Phavya-jfrog:

---
Wrap pipenv with the existing curation Post-Failure Run handler


itsmeleela (approved):

gauriy-tech (approved):

buildtools/cli.go
```diff
@@ -355,7 +355,10 @@ func GetCommands() []cli.Command {
 			SkipFlagParsing: true,
 			BashComplete:    corecommon.CreateBashCompletionFunc(),
 			Category:        buildToolsCategory,
-			Action:          PipenvCmd,
+			Action: func(c *cli.Context) error {
+				cmdName, _ := getCommandName(c.Args())
+				return securityCLI.WrapCmdWithCurationPostFailureRun(c, PipenvCmd, techutils.Pipenv, cmdName)
+			},
```
naveenku-jfrog, line 361: This wiring is correct and matches the pip/poetry pattern, but it's currently a no-op for the main case (jf pipenv install with build-info). WrapCmdWithCurationPostFailureRun → ShouldRunCurationAfterFailure only runs curation if the error is detected as a 403, via IsForbiddenOutput(PackageManager("pipenv"), ...) in build-info-go. That function has no "pipenv" case (verified at the pinned build-info-go commit 4f3044bf0940 used by jfrog-cli-security@v1.31.4), so it returns false and curation never runs. The install build-info path also never wraps ForbiddenError (only maven does), so the other two conditions in the gate don't catch it either.

This needs a companion change in build-info-go adding a pipenv case to IsForbiddenOutput (mirroring pip's http error 403), then a version bump through jfrog-cli-security → jfrog-cli. Without it, the wrapper adds overhead but never triggers the curation audit for pipenv. The same missing case also suppresses the user-facing curation-block message in GetMsgToUserForCurationBlock.

naveenku-jfrog (changes requested):

buildtools/cli.go
Phavya-jfrog: go.mod already points to 16a97012811d — the same commit you linked — which has the pipenv case in IsForbiddenOutput

Phavya-jfrog: go.mod already points to 16a97012811d — the same commit you linked — which has the pipenv case in IsForbiddenOutput

naveenku-jfrog (approved):

PR #3662, opened 2026-08-13
Parse first User-Agent product segment for product_id and version

sverdlov93:
## Summary
Fix Visibility `product_id` / `product_version` for long User-Agent strings from `jfrog-skills` / `jfrog-agent-hooks` by parsing the **first whitespace-delimited product segment** (`name/version`).

This preserves existing behavior everywhere else: HTTP and Call Home keep the full raw `JFROG_CLI_USER_AGENT`, and the existing AI suffix behavior is unchanged.

## Example
| Concern | Result |
|---|---|
| Visibility from `jfrog-skills/0.22.0 (trigger=skill; …) jfrog-cli-go/2.120.0` | `product_id=jfrog-skills`, `product_version=0.22.0` |
| HTTP / Call Home User-Agent | original full string preserved, with existing suffix behavior unchanged |

## Related
- dl-schema#287 (recognizes `client` / `model`)


yahavi (commented):
🤖 Review comment by Cursor AI

✅ All findings have been resolved.

- Uses `strings.Cut` for the first User-Agent segment
- Preserves the full raw User-Agent on HTTP / Call Home while Visibility uses only the first product segment
- Existing AI suffix behavior remains unchanged

⚠️ AI review supplements human review — always get a human reviewer too.

utils/cliutils/utils.go
```diff
@@ -58,10 +58,12 @@ func init() {
 	coreutils.SetClientAgentVersion(CliVersion)
 }
 
-// Splits the full agent name to its name and version.
-// The full agent name needs to be the agent name and version separated by a slash ('/').
-// If the full agent name doesn't include a version, then it's returned as the agent name and an empty string is returned as the agent version.
+// Splits the first User-Agent product token into its name and version.
+// If the token doesn't include a version, its version is returned empty.
 func splitAgentNameAndVersion(fullAgentName string) (string, string) {
+	if firstSpaceIndex := strings.Index(fullAgentName, " "); firstSpaceIndex != -1 {
```
yahavi: 🤖 **CLEANUP** — Prefer `strings.Cut` for the first-token trim

Equivalent and shorter:

```go
fullAgentName, _, _ = strings.Cut(fullAgentName, " ")
```

Same behavior when there is no space (`Cut` returns the original string as `before`).

⚠️ AI review supplements human review — always get a human reviewer too.

yahavi: 🤖 **BUG** — Rich UA lost on HTTP / Call Home wire

Truncating at the first space fixes Visibility `product_id` / `product_version`, but `coreutils.GetCliUserAgent()` rebuilds only `name/version`. So for:

`jfrog-skills/0.22.0 (trigger=skill; …) jfrog-cli-go/2.120.0`

- **Before** (LastIndex quirk): HTTP/Call Home kept the full rich string
- **After**: wire becomes `jfrog-skills/0.22.0` only — `trigger=` and trailing `jfrog-cli-go/…` are dropped

`main.go` sets the client UA via `GetCliUserAgentWithAgent()` → `GetCliUserAgent()`, and Call Home uses `GetCliUserAgent()` too. Commit `b701162` preserved raw env with `cliUserAgentRaw`; `90fc905` removed it. The PR notes still say skill-vs-hook stays a Coralogix `trigger=` parens signal — that no longer reaches the wire.

**Fix:** Restore the raw-base passthrough (and ai-* dedup) from `b701162`, plus `TestGetCliUserAgentWithAgentPreservesRichRawUA`.

⚠️ AI review supplements human review — always get a human reviewer too.

sverdlov93: ✅ Fixed — restored `cliUserAgentRaw` passthrough (+ ai-* dedup) so HTTP/Call Home keep the full rich UA (including `trigger=`), while Visibility product_id/product_version still come from the first product token only. Added `TestGetCliUserAgentWithAgentPreservesRichRawUA` and `TestGetCliUserAgentWithAgentNoDuplicateAiAgent`.

sverdlov93: ✅ Fixed — first-token trim now uses `strings.Cut`.

yahavi (approved):

PR #3616, opened 2026-07-24
RTECO-1574 - RTECO-1574 - Implementation of Nuget Support for client -

bhanurp:
Generated by workbench RTECO-1574.

## RTECO-1574 - Implementation of Nuget Support for client -

**Source:** https://jfrog-int.atlassian.net/browse/RTECO-1574
- Start working on Nuget V2, V3 support for Nuget package manager in jfrog cli. 
- Make sure PRs are manageable and code is written in a way it can be reused across the clients 
- Add support for nugetV3 and nugetV2
- refer to [unsupported block: inlineCard]
## Considerations while implementation
- Use JFROG_CLI_NATIVE_IMPLEMENTATION support since nuget is already supported when this is set it routes via native package manager(flexpack).
- BuildInfo collection in build-info-go
- check sum calculation
- Original Deployment Repository calculation
- Requested By calculation for dependencies
- Easy authentication using nuget’s credentials like mentioned in above Atlassian wiki.
- Build Info collection for both dependencies and artifacts for all the commands eligible listed in above wiki.
- Set properties clearly on artifacts published.
- Make sure the buildinfo is figuring out and showing show in tree in build info section when published to artifactory.
## Not to implement
- nuget-config support is not required since this is wrapper on native binary.
**Parent:** RTECO-395
**Components:** jfrog-cli-nuget

---
Evidence: see Artifactory bundle for RTECO-1574.

Jira: https://jfrog-int.atlassian.net/browse/RTECO-1574

Issue #3377, opened 2026-03-02
Can't publish the .NET/NuGet packages with build information via 'jf dotnet nuget push' command


yanxiuwu:
### Describe the bug

If we use the command:
`jf dotnet nuget push *.nupkg --source "MyArtifact" --build-name="xxx" --build-number="x"` to push the .NET packages to the Artifactory, we can't find the build information under **Builds**,
But use the command:
`jf rt u "*.nupkg" $nuget-repo-name --build-name="sample_project_build_info" --build-number="1.0.1"` can push the .NET packages to the Artifactory with the build information

### Current behavior

We use the command:
`jf dotnet nuget push *.nupkg --source "MyArtifact" --build-name="sample_project_build_info" --build-number="1.0.1"` to push the .NET packages to the Artifactory, we can't find the build information under **Builds**,
But use the command:
`jf rt u "*.nupkg" $nuget-repo-name --build-name="sample_project_build_info" --build-number="1.0.1"` can push the .NET packages to the Artifactory with the build information

### Reproduction steps

jf dotnet-config --server-id-resolve xxx --repo-resolve xxxx
jf dotnet restore --build-name=xxx --build-number=x
jf dotnet pack --configuration Release --output ./ --build-name=xxx --build-number=x
jf dotnet nuget push *.nupkg --source "MyArtifact" --build-name="xxx" --build-number="x" 
jf rt bp xxx x

### Expected behavior

Can use the command `jf dotnet nuget push *.nupkg --source "MyArtifact" --build-name="sample_project_build_info" --build-number="1.0.1"` to publish the packages with build information

### JFrog CLI version

jf version 2.93.0

### Operating system type and version

macOS Tahoe 26.3

### JFrog Artifactory version

7.133.10

SierraNL:
I have similar issues on Linux, so it's not macOS specific.

Looking at this issue, It would be helpful to at least document in what step, what build-info is gathered. Because with dotnet there are so many different ways of doing things, and you often end up without build info. Without knowing if you don't use the rights steps or hit an issue. Extra debugging info around this would be great also.

oronf-mda:
Support for `jf dotnet nuget push` is important because `jt rt u` cannot upload symbol packages.

oronf-mda:
As of version 2.124.0 this is still an issue

PR #3664, opened 2026-08-14
RTECO-1574 - Implementation of Nuget Support for client -

bhanurp:
Generated by workbench RTECO-1574.

## RTECO-1574 - Implementation of Nuget Support for client -

- Start working on Nuget V2, V3 support for Nuget package manager in jfrog cli.
- Make sure PRs are manageable and code is written in a way it can be reused across the clients
- Add support for nugetV3 and nugetV2
- refer to [unsupported block: inlineCard]
## Considerations while implementation
- Use JFROG_RUN_NATIVE support since nuget is already supported when this is set it routes via native package manager(flexpack).
- BuildInfo collection in build-info-go
- check sum calculation
- Original Deployment Repository calculation
- Requested By calculation for dependencies
- Easy authentication using nuget's credentials like mentioned in above Atlassian wiki.
- Build Info collection for both dependencies and artifacts for all the commands eligible listed in above wiki.
- Set properties clearly on artifacts published.
- Make sure the buildinfo is figuring out and showing show in tree in build info section when published to artifactory.
## Not to implement
- nuget-config support is not required since this is wrapper on native binary.
**Parent:** RTECO-395
**Components:** jfrog-cli-nuget

---
Evidence: see Artifactory bundle for RTECO-1574.

Jira: https://jfrog-int.atlassian.net/browse/RTECO-1574

---
Supersedes #3616, opened from a branch pushed directly to this repo (rather than a fork) so that `pull_request_target`-gated workflows (e.g. apt tests) resolve their workflow YAML from the correct ref.


SierraNL:
Is this a fix for issue #3377 ?

testdata/gradle/projectwithplugin/build/reports/problems/problems-report.html
```diff
@@ -0,0 +1,659 @@
+<!DOCTYPE html>
```
agrasth: We can remove this file from PR.

buildtools/cli.go
```diff
@@ -2162,6 +2186,96 @@ func ConanCmd(c *cli.Context) error {
...
+	// push itself on a critical vulnerability, as Maven's --scan does via
+	// commandsUtils.ConditionalUploadScanFunc) isn't wired for NuGet FlexPack yet - same
+	// documented gap as the curation hook (scenario 11) - so for now this only strips the flag
+	// rather than acting on it.
+	xrayScan, err := cliutils.ExtractBoolFlagFromArgs(&filteredArgs, "scan")
+	if err != nil {
+		return err
+	}
+	if xrayScan {
+		log.Debug("'--scan' was passed to 'jf nuget push' but conditional-upload Xray scanning is not yet wired for NuGet FlexPack; the flag is accepted and stripped, not acted on.")
```
agrasth: `--scan` is *silently stripped*, not honored. As written, `jf nuget push --scan` returns success without doing any scanning. That's worse than not supporting the flag — better to fail loudly until the wiring is in place:

```go
if xrayScan {
    return errors.New("--scan is not yet supported for NuGet FlexPack; omit --scan or use the legacy path")
}
```

Then wire it in a follow-up. Silent no-op flags are a well-known source of "we thought we were secure" bugs.

utils/tests/proxy/server/server.go
```diff
@@ -47,16 +47,20 @@ func getReverseProxyHandler(targetUrl string) (*httputil.ReverseProxy, error) {
...
-		req.Host = origHost
-		req.URL.Scheme = target.Scheme
-	}
+	origScheme := target.Scheme
 	tr := &http.Transport{
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 	}
 	proxyErrLogger := log.New(os.Stdout, "PROXY-LOGGER", log.Ldate|log.Ltime|log.Lshortfile)
-	p := &httputil.ReverseProxy{Director: d, Transport: tr, ErrorLog: proxyErrLogger}
+	p := &httputil.ReverseProxy{
```
agrasth: Migrates `Director` → `Rewrite` on `httputil.ReverseProxy`. Correct move (Director is deprecated as of Go 1.20), but this proxy is shared by many test suites (Docker, Podman, Maven, ...). Please confirm those all still pass and call out the migration in the PR description so reviewers know the blast radius extends beyond NuGet.

agrasth:
Can you pls check what is the env var for native mode, PR decp says as `JFROG_CLI_NATIVE_IMPLEMENTATION`, it should be `JFROG_RUN_NATIVE`.

bhanurp:
> Is this a fix for issue #3377 ?

@SierraNL yes this will be fixed as well, but in general this PR and other related PRs covers complete implementation for nuget v2 and v3 support.

testdata/gradle/projectwithplugin/build/reports/problems/problems-report.html
bhanurp: Fixed removed from the PR.

buildtools/cli.go
bhanurp: Fixed `--scan` now returns a hard error (`flag --scan is not supported for native NuGet/dotnet commands`) instead of silently stripping the flag.

utils/tests/proxy/server/server.go
bhanurp: Good catch, this change only affects the proxy test server used in integration tests (not production traffic). The `Rewrite` hook receives the same `*httputil.ProxyRequest` and sets `Out.Header` equivalently to the old `Director`. Scope is limited to `utils/tests/proxy/server/server.go`; no production proxy path is affected.

bhanurp:
> Can you pls check what is the env var for native mode, PR decp says as `JFROG_CLI_NATIVE_IMPLEMENTATION`, it should be `JFROG_RUN_NATIVE`.

yes it is JFROG_RUN_NATIVE across the codebase.

buildtools/cli.go
```diff
@@ -1030,11 +1033,22 @@ func NugetCmd(c *cli.Context) error {
...
 	}
 
-	configFilePath, err := getProjectConfigPathOrThrow(project.Nuget, "nuget", "nuget-config")
+	configFilePath, configExists, err := project.GetProjectConfFilePath(project.Nuget)
 	if err != nil {
 		return err
 	}
 
+	// FlexPack bypasses all config file requirements (only when no config exists)
+	if artutils.ShouldRunNative(configFilePath) && !configExists {
```
fluxxBot, line 1057: let's add debug comment when run native is invoked

fluxxBot (approved):

fluxxBot (approved):

PR #3676, opened 2026-08-20
remove -k (--insecure) flag from curl

TheSkibb:
- [X] All [tests](https://github.com/jfrog/jfrog-cli/blob/master/CONTRIBUTING.md#tests) have passed. If this feature is not already covered by the tests, new tests have been added.
- [X] The pull request is targeting the `master` branch.
- [X] The code has been validated to compile successfully by running `go vet ./...`.
- [X] The code has been formatted properly using `go fmt ./...
**(no go code in commit)**
---

the -k (--insecure) flags skips the verification of whether or not the transfer is actually secure 

see https://daniel.haxx.se/blog/2020/01/27/curl-ootw-k-insecure/ on why --insecure really means insecure

since the files are served from jfrogs domain it is in the users interest to know that the certificate is actually validated


TheSkibb:

PR #3681, opened 2026-08-25
Fix npmnpmrcproject fixture: scope legacy auth keys to the registry

agrasth:
## Summary
- npm 12 hard-rejects the bare `email` and `_auth` npmrc auth config keys with `ERR_INVALID_AUTH`, which was failing `TestNpmNativeSyntax/npm_i_with_npmrc_project` (and the equivalent legacy-syntax test) for every npm 12.x version in our pm-version-monitor compatibility tracking — the very first `npm install` call in the test errored out before jfrog-cli's own npm command logic ever ran.
- This is a test-fixture staleness issue, not a jfrog-cli npm command bug: `testdata/npm/npmnpmrcproject/.npmrc` still used the pre-npm-12 unscoped auth key format.
- Scope both keys to the registry host, matching the format the fixture's existing `_authToken` line already uses (and exactly what npm's own error message names as the fix).

## Test plan
- [x] Verified locally: `npm install --dry-run` against the original fixture with npm 12.0.0 reproduces `ERR_INVALID_AUTH: Invalid auth configuration found: 'email' must be renamed to '//<registry>/:email' in project config`
- [x] Verified the fixed fixture passes config validation with npm 12.0.0 (fails only on `ENOTFOUND` for the fixture's placeholder `NO-NO-REPO` host, which is expected — the real test redirects to a live Artifactory instance)
- [x] `go vet ./...` and `gofmt` — no changes needed, fixture-only diff


fluxxBot (changes requested):
I don't think this changes are valid, let's re look at problem and see how can be this solved

PR #3690, opened 2026-09-01
Add username to atc output

CharlieAtSchneider:
This PR adds username to `jf atc` output. Since it isn't appearing in the reply to the API call, it is instead manually appended to the output in code.

Please tell me if you think the idea is sound. If so, I can fix my issue in the Caveats section below.

Examples:
`jf atc`:
```
{
  "access_token": "<TOKEN>",
  "expires_in": 15552000,
  "scope": "applied-permissions/user",
  "token_id": "<ID>",
  "token_type": "Bearer",
  "username": "<MY_USERNAME>"
}
```
`jf atc --format=table`
```
FIELD         VALUE
access_token  <TOKEN>
token_id      <ID>
expires_in    1.5552e+07
scope         applied-permissions/user
token_type    Bearer
username      <MY_USERNAME>

```

## Use case
If you can't run `jf docker` commands, (in our case due to VM shenanigans) you may need to have a username together with the token to log in to Docker. This addition to `jf atc` would fix this.

## Caveats:
* I am not a Go developer, and I created this code by looking in the source code for reference. Please correct me if I made any simple mistakes.
* I'm not super fond of the idea of running `json.Unmarshal()`, add a username field to the struct, turn it back with `json.Marshal()` for it to just turn back with `json.Unmarshal()` later in the code in other functions that I didn't touch. What do you think - should this be avoided with the help of some refactoring?

## Tests, formatting etc

All code written by a human


CharlieAtSchneider:

CharlieAtSchneider:
Any inputs or thoughts on the general idea?

CharlieAtSchneider:
Still awaiting response

Issue #156, opened 2018-04-18
Unsupported protocol scheme "" error using default url


johngmyers:
Jfrog version 1.14.0

Attempting to perform commands using the url of the default server-id results in an `Unsupported protocol scheme ""` error:

```sh
~ $ rm .jfrog/jfrog-cli.conf 
~ $ jfrog rt config default --interactive=false --url=http://default.invalid/artifactory
~ $ jfrog rt use default
~ $ jfrog rt copy --user=user --password=hunter2 source/foo dest
[Info] Searching artifacts...
[Error] Post api/search/aql: unsupported protocol scheme ""
{
  "status": "success",
  "totals": {
    "success": 0,
    "failure": 0
  }
}
```

Using an explicit `--set` switch gives the expected result:

```sh
~ $ jfrog rt copy --user=user --password=hunter2 --url=http://default.invalid/artifactory source/foo dest
[Info] Searching artifacts...
[Error] Post http://default.invalid/artifactory/api/search/aql: dial tcp: lookup default.invalid on 10.0.2.3:53: no such host
{
  "status": "success",
  "totals": {
    "success": 0,
    "failure": 0
  }
}
```

jackbezalel:
I have this issue as well.. Seems to me it is a bug in this latest CLI version..
Maybe try and use a previous version of the CLI, or the REST API instead

arbourd:
Wrap the url in `""`.

eyalbe4:
@johngmyers and @jackbezalel,
Do you still see this issue? In case you do, can you please try upgrading to version 1.16.0?
Please note that due to security concerns, JFrog CLI no longer allows sending the *--user* and *--password* options without sending the *--url* option.

nicevo:
@eyalbe4 I am still seeing this issue today. I am on the latest homebrew version. 
 
circleci version
0.1.15338+850f9ac (homebrew)

yahavi:
@nicevo 
Make sure to properly provide the URL in the config command.
Option 1 - Provide JFrog platform URL:
```bash
jfrog config add --url=https://<base-platform-url> --username=<username> --password=<password> --interactive=false
```
Option 2 - Provide Artifactory URL:
```bash
jfrog config add --artifactory-url=https://<base-platform-url>/artifactory --username=<username> --password=<password> --interactive=false
```

Please let me know if that helped.

approximatenumber:
[...]

Issue #151, opened 2018-04-08
Return only path property


mazzy89:
Why JFrog CLI returns only `path` property?

eyalbe4:
@mazzy89,
You probably refer to the output of the *jfrog rt search* command.
We've recently merged [this](https://github.com/JFrogDev/jfrog-cli-go/pull/104) pull request, which adds the artifact properties to the command's response.

mazzy89:
Yes I mean that. I see so it will be available in some next releases at the moment or just from building the master. Thank you @eyalbe4

Issue #150, opened 2018-04-05
"Incorrect Usage" error still returns 0 return-code


dsoprea:
```
Incorrect Usage.

NAME:
   jfrog rt upload - Upload files.

USAGE:
   jfrog rt upload [command options] [arguments...]

OPTIONS:
   --url 		[Optional] Artifactory URL.
   --user 		[Optional] Artifactory username.
   --password 		[Optional] Artifactory password.
...
   --explode		[Default: false] Set to true to extract an archive after it is deployed to Artifactory.
   --symlinks		[Default: false] Set to true to preserve symbolic links structure in Artifactory.
   --include-dirs	[Default: false] Set to true if you'd like to also apply the source path pattern for directories and not just for files.
   
~$ echo $?
0
```

I realize that we're using a third-party package for argument processing, but I can't figure out where this message even comes from in order to request/submit a fix for it and then port it here.

```
src$ grep "Incorrect usage" -R .
./github.com/jfrogdev/jfrog-cli-go/vendor/github.com/codegangsta/cli/app.go:	// If this function is not set, the "Incorrect usage" is displayed and the execution is interrupted.
./github.com/jfrogdev/jfrog-cli-go/vendor/github.com/codegangsta/cli/command.go:	// If this function is not set, the "Incorrect usage" is displayed and the execution is interrupted.
```

eyalbe4:
@dsoprea,
Are you using the latest JFrog CLI version? It looks like this issue is fixed in *1.14.0*:
```
eyalb-mac:helm eyalb$ 
eyalb-mac:helm eyalb$ jfrog -v
jfrog version 1.14.0
eyalb-mac:helm eyalb$ jfrog rt u abc
[Error] Wrong number of arguments. You can read the documentation at https://www.jfrog.com/confluence/display/CLI/JFrog+CLI

Name:
	jfrog rt upload - Upload files.

Usage:
	jfrog rt u [command options] <source pattern> <target pattern>
	jfrog rt u --spec=<File Spec path> [command options]

Arguments:
	source pattern
		Specifies the local file system path to artifacts which should be uploaded to Artifactory.
		You can specify multiple artifacts by using wildcards or a regular expression as designated by the --regexp command option.
		If you have specified that you are using regular expressions, then the first one used in the argument must be enclosed in parenthesis.

	target pattern
		Specifies the target path in Artifactory in the following format: <repository name>/<repository path>.
[...]

Issue #145, opened 2018-03-29
Error should be throw when a 403 is returned from the server


shawnbutts:
What i'm seeing (via strace) with `jfrog rt dl --url http://privateserver:8081/artifactory --threads 5 /a.spec` is that a 403 is being returned the server and jfrog-cli-go just hangs with `[Info] Searching items to download...`

elhigu:
I was experiencing the same kind of hanging and `jfrog rt search` command actually in my case gave more information why (some authentication problem). Anyways my workaround was to use jfrog client version 1.13.1.

yahavi:
@shawnbutts & @elhigu 
Thanks for reporting this issue.
We fix this issue in the latest release v1.16.0. Feel free to upgrade.
We'll appreciate you feedback for that.

PR #3699, opened 2026-09-06
RTECO-1782: JFROG_RUN_NATIVE wins over project config, fix dotnet help, add FlexPack tests

bhanurp:
## What

Two commits: a behaviour fix in the FlexPack gate with the help text it depends on, and the dotnet FlexPack integration suite.

---

## 1. `JFROG_RUN_NATIVE` now wins over a project config file

The gate was:

```go
if artutils.ShouldRunNative(configFilePath) && !configExists {
```

so **any** `.jfrog/projects/{dotnet,nuget}.yaml` in a project forced the legacy path even with `JFROG_RUN_NATIVE=true`. Nothing was logged.

Worse, the legacy path doesn't recognise the native-only flags, so it forwarded them to MSBuild:

```
MSBUILD : error MSB1001: Unknown switch.
Switch: --repo-resolve
```

— which names an MSBuild flag and gives no hint that a YAML file two directories down is the cause. And if the stale config happened to be *valid*, the build silently resolved from whatever repository that file named rather than the `--repo-resolve` on the command line. Wrong-repo resolution, no diagnostic.

Now the env var takes precedence and the config is reported and ignored:

```
[Warn] JFROG_RUN_NATIVE=true, so the dotnet configuration at ".../dotnet.yaml" is being
       ignored and the command runs in native (FlexPack) mode. Unset JFROG_RUN_NATIVE to
       use the legacy 'jf dotnet-config' path.
```

Applied to **both** `DotnetCmd` and `NugetCmd` — identical gate, identical failure.

Regression-verified, all three paths:

| Setup | Behaviour |
|---|---|
| `JFROG_RUN_NATIVE=true` + config present | warns, runs FlexPack ✅ |
| env unset + config present | legacy path still works ✅ |
| env unset + no config | original "run `jf dotnet-config` first" error ✅ |

### The help text was the other half

`jf dotnet --help` listed `jf dotnet-config` as a prerequisite — *"must be run first"*. So a user following the CLI's own documentation created exactly the file that disabled FlexPack. That command is out of scope for FlexPack per the spec, yet remained the documented happy path. Now described as optional under `JFROG_RUN_NATIVE`, for both `dotnet` and `nuget`.

Also corrected the sub-command list, which claimed `(restore, build, pack, push)`. **There is no `jf dotnet push`** — the real command is the two-token `jf dotnet nuget push`, fully implemented in `getNugetCommandName` but documented nowhere. Now: restore, build, publish, pack, add, and `nuget push`, with an example and an explicit note.

This closes two open bug-hunt tickets about `-h` not showing supported args and `jf dotnet nuget` being undocumented.

---

## 2. dotnet FlexPack integration tests

Adds `dotnet_native_test.go`. That combination had **no coverage** — the empty quadrant:

| File | Toolchain | Path |
|---|---|---|
| `nuget_test.go` | nuget.exe + dotnet | legacy |
| `nuget_native_test.go` | nuget.exe | FlexPack |
| **`dotnet_native_test.go`** | **dotnet** | **FlexPack** ← new |

Derived from the Confluence test plan (RTFACT 2729476103). **114 tests against its 186 scenarios: all 40 P0, all 106 P1, 38/40 P2.** Every test names the scenarios it covers, so coverage is auditable against the plan rather than asserted.

26 are `t.Skip` with a stated reason, in three groups:
- **Known product gaps** — curation-on-failure hook not wired for dotnet; `--scan` accepted on push but stripped
- **Missing fixtures** — `.fsproj` / `.vbproj` / `.slnx`, signed packages, >100 MB packages
- **Infrastructure** — promotion, Xray, release bundles, CI simulation, self-signed TLS, proxy. Each skip names the `nuget_native_test.go` helper to reuse when porting.

Infrastructure mirrors `nuget_native_test.go` and **reuses** its shared helpers rather than duplicating them. Runs under the existing `-test.nuget` flag, which is what provisions the NuGet repositories these tests share — there is no separate dotnet flag. `nuget_native_test.go` is untouched.

---

## ⚠️ Spec divergences, asserted as-implemented

Five P0 scenarios contradict the implementation. Following the precedent already set in `nuget_native_test.go`, the tests pin **current behaviour**, and the file header lists each divergence with both readings so a change in either direction shows up as a failure:

| Scenario | Plan says | Code does |
|---|---|---|
| #145 / #156 | no temp `nuget.config` is written | one **is** written (sources only) |
| #150 | JFrog creds used only for post-push stamping | also used for resolve |
| #151 | creds **not** exported to child env | they **are**, via `NuGetPackageSourceCredentials_*` |
| #13 | `<repo>/<Name>/<Version>/<file>.nupkg` | lands **flat** at `<repo>/<file>.nupkg` |

**These need reconciling with the spec owner** — either the spec or the implementation is wrong, and this PR doesn't decide which.

## Testing

- `gofmt`, `go build ./...`, `go vet`, `golangci-lint` — all clean
- Test package compiles; verified end-to-end against a live Artifactory across 8 project types

> **Notes:** `gosec` could not be run (`internal error: package "embed" without types` under this Go toolchain) — unverified rather than passing. The new tests are compile- and vet-verified but have **not** yet been run against a live Artifactory, so expect some to need adjustment on first CI run.

## Merge order

Last of three RTECO-1782 PRs. Both dependencies must be merged **and bumped** first — `TestDotnetFlexPackRequestedByHasNoRedundantPaths` asserts the build-info-go dedupe behaviour.

1. jfrog/build-info-go#422
2. jfrog/jfrog-cli-artifactory#551
3. **jfrog-cli** ← this PR

### DIFF
--- a/buildtools/cli.go
+++ b/buildtools/cli.go
@@ -1040,6 +1040,30 @@ func extractPnpmOptionsFromArgs(args []string) (serverDetails *coreConfig.Server
 	return serverDetails, cleanArgs, buildConfig, nil
 }
 
+// shouldRunNuGetFlexPack reports whether the FlexPack (native) path should handle a
+// 'jf nuget' / 'jf dotnet' invocation.
+//
+// JFROG_RUN_NATIVE=true takes precedence over a per-project configuration file. Previously the
+// gate was `ShouldRunNative(configFilePath) && !configExists`, so any leftover
+// .jfrog/projects/{nuget,dotnet}.yaml silently forced the legacy path even with the
+// environment variable set. That was invisible to the user, and because the legacy path does
+// not recognise the native-only flags it forwarded them to MSBuild, surfacing as an opaque
+// "MSBUILD : error MSB1001: Unknown switch --repo-resolve". The config file is now reported
+// and ignored instead.
+//
+// configFilePath is only used for the warning message; pass configExists to say whether one
+// was found. pmName names the package manager for the 'jf <pm>-config' hint.
+func shouldRunNuGetFlexPack(configFilePath string, configExists bool, pmName string) bool {
+	// ShouldRunNative("") is IsFlexPackEnabled() with no config-path condition attached.
+	if !artutils.ShouldRunNative("") {
+		return false
+	}
+	if configExists {
+		log.Warn(fmt.Sprintf("JFROG_RUN_NATIVE=true, so the %s configuration at %q is being ignored and the command runs in native (FlexPack) mode. Unset JFROG_RUN_NATIVE to use the legacy 'jf %s-config' path.", pmName, configFilePath, pmName))
+	}
+	return true
+}
+
 func NugetCmd(c *cli.Context) error {
 	if show, err := cliutils.ShowCmdHelpIfNeeded(c, c.Args()); show || err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


bhanurp:
Superseded by #3703, which is the same work on a branch in this repository rather than a fork. A PR's head branch cannot be changed after creation, so reopening was the only way to move it.

PR #3701, opened 2026-09-07
Add GitHub Actions e2e dispatch workflow (JFrog Pipelines EOL migration)

agrasth:
Ports `jfrog_artifactory_ecomatrix_cli_e2e` from `.jfrog-pipelines/pipelines.yml` to a `workflow_dispatch` GitHub Actions workflow, as part of migrating off JFrog Pipelines (EOL).

**This is a best-effort port, not a byte-for-byte translation — please review before merging.**

The pipeline is really two things: (1) a GitHub Actions dispatch-and-poll harness that mints an OAuth token and triggers/monitors ~16 test workflows in `jfrog-cli-workflows`, and (2) two calls out to a Jenkins job (`tools/platform/environment_setup_gen2` and `tools/platform/environment_operate`) to stand up and tear down an ephemeral Artifactory.

- **Part 1 is carried over near-verbatim.** The original bash was already host-agnostic curl/jq against a configurable `GITHUB_API_URL` (defaults to `https://github.jfrog.info/api/v3`) — copied essentially unchanged into a `run:` step.
- **Part 2 is reimplemented, not ported**, because JFrog Pipelines' native `Jenkins` step type doesn't expose *how* it talks to Jenkins in the YAML — there's nothing to literally copy. I implemented it against the generic Jenkins remote-build API: fetch a CSRF crumb, POST to `buildWithParameters`, follow the returned queue-item until it has a build, then poll that build until it completes. **This needs verification against your actual Jenkins instance** — in particular whether the job path `tools/platform/environment_setup_gen2` resolves the same way via `/job/tools/job/platform/job/environment_setup_gen2/` as it does through the native integration, and whether crumb-based CSRF is even enabled there.
- **Network reachability is unverified**: this job needs outbound access to `https://github.jfrog.info` (JFrog's GHE instance) and to the Jenkins host from wherever it runs. If either is only reachable from inside JFrog's network, this workflow needs a self-hosted runner rather than `ubuntu-latest` — I don't have visibility into that from the YAML alone.

**Secrets required** (repo Settings → Secrets and variables → Actions):
- `JFROG_CLI_TESTS_MASTER_KEY`, `JFROG_CLI_TESTS_USERNAME`, `JFROG_CLI_TESTS_PASSWORD`, `JFROG_CLI_TESTS_TOKEN` — from the `jfrog_cli_tests` integration
- `GITHUB_DISPATCH_TOKEN` — from the `github_dispatch` integration (must be a fine-grained PAT or GitHub App token; the JFROG org on github.jfrog.info rejects classic PATs)
- `JFROG_CLI_GH_TOKEN` — fallback token from `jfrog_cli_gh`, used only if `GITHUB_DISPATCH_TOKEN` is unset
- `JENKINS_ENTPLUS_RT_URL`, `JENKINS_ENTPLUS_RT_USER`, `JENKINS_ENTPLUS_RT_TOKEN` — from the `jenkins_entplus_rt` integration

Not run yet — needs the secrets above added first, and the Jenkins-triggering steps verified against the real instance.

### DIFF
--- a/.github/scripts/e2e-dispatch.sh
+++ b/.github/scripts/e2e-dispatch.sh
@@ -1,0 +1,246 @@
+#!/usr/bin/env bash
+# Ported from .jfrog-pipelines/pipelines.yml (JFrog Pipelines is EOL).
+#
+# Runnable both from the "E2E Dispatch" GitHub Actions workflow and from a
+# local shell. To run locally, export the env vars below (same names the
+# workflow injects from secrets/inputs) and run this script directly from a
+# checkout of jfrog/jfrog-cli.
+#
+# Required:
+#   JENKINS_URL, JENKINS_USER, JENKINS_TOKEN         (jenkins_entplus_rt integration)
+#   MASTER_KEY, JFROG_ADMIN_USERNAME,
+#   JFROG_ADMIN_PASSWORD, JFROG_ADMIN_TOKEN          (jfrog_cli_tests integration)
+#   GITHUB_DISPATCH_TOKEN or JFROG_CLI_GH_TOKEN       (github_dispatch / jfrog_cli_gh integration;
+#                                                      fine-grained PAT or GitHub App token —
+#                                                      github.jfrog.info rejects classic PATs)
+#
+# Optional (defaults shown):
+#   RT_VERSION=                    SKIP_ENV_SETUP=false      ARTIFACTORY_URL=
+#   GITHUB_API_URL=https://github.jfrog.info/api/v3
+#   GITHUB_WORKFLOWS_REPO=JFROG/jfrog-cli-workflows          GITHUB_WORKFLOWS_REF=master
+#   JFROG_CLI_GITHUB_REPO=jfrog/jfrog-cli                    JFROG_CLI_GITHUB_REF=<current HEAD sha>
+#   GHE_ACTIONS_RUNNER=artifactory-dind-amd-scale-set
+#   LOGS_TO_KIBANA=true             DEPLOYMENT_SIZING=common
+#   MAX_RUN_RETRIES=2               MAX_WAIT_SECONDS=14400
+#
+# NOT independently verified against a real Jenkins instance: the provision/
+# teardown steps call the generic Jenkins remote-build API (crumb + queue
+# polling) because JFrog Pipelines' native "Jenkins step type" doesn't expose
+# how it itself talks to Jenkins — there was nothing literal to port here.
+
+set -eo pipefail
+
+fail() { echo "ERROR: $1" >&2; exit 1; }
+
+SKIP_ENV_SETUP="${SKIP_ENV_SETUP:-false}"
+GITHUB_API_URL="${GITHUB_API_URL:-https://github.jfrog.info/api/v3}"
+GITHUB_WORKFLOWS_REPO="${GITHUB_WORKFLOWS_REPO:-JFROG/jfrog-cli-workflows}"
+GITHUB_WORKFLOWS_REF="${GITHUB_WORKFLOWS_REF:-master}"
+JFROG_CLI_GITHUB_REPO="${JFROG_CLI_GITHUB_REPO:-jfrog/jfrog-cli}"
+JFROG_CLI_GITHUB_REF="${JFROG_CLI_GITHUB_REF:-$(git rev-parse HEAD 2>/dev/null || echo "")}"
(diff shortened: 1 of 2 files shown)


.github/scripts/e2e-dispatch.sh
```diff
@@ -0,0 +1,246 @@
...
+    sleep 15
+  done
+  result=$(echo "$build_json" | jq -r '.result')
+  echo "environment_setup_gen2 result: ${result}"
+  [[ "${result}" == "SUCCESS" ]] || fail "environment provisioning failed"
+fi
+
+# --- Mint an OAuth token against the (now ready) environment ---
+echo "Installing JFrog CLI 2.54.0"
+curl -fL https://getcli.jfrog.io/v2 | sh -s 2.54.0
```
bhanurp, line 126: why not latest version?

.github/scripts/e2e-dispatch.sh
```diff
@@ -0,0 +1,246 @@
...
+# Optional (defaults shown):
+#   RT_VERSION=                    SKIP_ENV_SETUP=false      ARTIFACTORY_URL=
+#   GITHUB_API_URL=https://github.jfrog.info/api/v3
+#   GITHUB_WORKFLOWS_REPO=JFROG/jfrog-cli-workflows          GITHUB_WORKFLOWS_REF=master
+#   JFROG_CLI_GITHUB_REPO=jfrog/jfrog-cli                    JFROG_CLI_GITHUB_REF=<current HEAD sha>
+#   GHE_ACTIONS_RUNNER=artifactory-dind-amd-scale-set
+#   LOGS_TO_KIBANA=true             DEPLOYMENT_SIZING=common
+#   MAX_RUN_RETRIES=2               MAX_WAIT_SECONDS=14400
+#
+# NOT independently verified against a real Jenkins instance: the provision/
```
bhanurp, line 26: you can create a test job verify to address any hidden scenarios

.github/workflows/e2e-dispatch.yml
```diff
@@ -0,0 +1,67 @@
...
+      JFROG_CLI_GH_TOKEN: ${{ secrets.JFROG_CLI_GH_TOKEN }}
+      JENKINS_URL: ${{ secrets.JENKINS_ENTPLUS_RT_URL }}
+      JENKINS_USER: ${{ secrets.JENKINS_ENTPLUS_RT_USER }}
+      JENKINS_TOKEN: ${{ secrets.JENKINS_ENTPLUS_RT_TOKEN }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Run e2e dispatch
+        run: .github/scripts/e2e-dispatch.sh
```
bhanurp, line 67: instead of using the script, cant we invoke using github actions itself and check their status or using gh cli.

.github/scripts/e2e-dispatch.sh
```diff
@@ -0,0 +1,246 @@
...
+}
+[[ "${SKIP_ENV_SETUP}" == "true" ]] || trap teardown EXIT
+
+# --- Jenkins step: tools/platform/environment_setup_gen2 ---
+# VERIFY: confirm this job path resolves the same way against your Jenkins
+# instance as it did through JFrog Pipelines' native Jenkins integration.
+if [[ "${SKIP_ENV_SETUP}" != "true" ]]; then
+  echo "Provisioning ephemeral Artifactory..."
+  job_path="job/tools/job/platform/job/environment_setup_gen2"
+  crumb_json=$(curl -sS -u "${JENKINS_USER}:${JENKINS_TOKEN}" "${JENKINS_URL}/crumbIssuer/api/json" || echo '{}')
```
bhanurp, line 87: use an existing github action or write reusable github action to reuse it across the workflows instead of simple script which can trigger jenkins job that way you can maintain and scale triggering jenkins jobs

.github/scripts/e2e-dispatch.sh
```diff
@@ -0,0 +1,246 @@
+#!/usr/bin/env bash
+# Ported from .jfrog-pipelines/pipelines.yml (JFrog Pipelines is EOL).
+#
+# Runnable both from the "E2E Dispatch" GitHub Actions workflow and from a
+# local shell. To run locally, export the env vars below (same names the
```
bhanurp, line 5: > To run locally
why do we need to run locally? Isnt it from CI 100% of the times?

bhanurp (changes requested):

PR #3704, opened 2026-09-08
Verify checksum in official CLI install scripts

ehl-jf:
The five official install scripts (`build/getcli/jf.sh`, `build/getcli/jfrog.sh`, `build/installcli/jf.sh`, `build/installcli/jfrog.sh`, `build/setupcli/jf.sh`) now verify the downloaded `jf`/`jfrog` binary's SHA256 before ever `chmod +x`-ing or installing it onto PATH, and no longer skip TLS verification (`curl -k` removed).

The checksum is fetched from the per-binary `.sha256` sidecar JGC-542 publishes at the exact same path as the binary (e.g. `jfrog-cli-linux-amd64/jf.sha256`) — verified end-to-end against the real, published `releases.jfrog.io` endpoint. The install aborts (deleting the partial binary) on a checksum mismatch, an unreachable/missing sidecar, or an empty/malformed one, with failure messages naming both the binary and checksum URLs and both hashes where relevant.

Adds a dependency-free test driver (`build/scripts_test/run_tests.sh`, 17 cases) that builds fixtures on the fly matching whatever OS/arch the machine actually is — real subprocess execution against `file://` URLs, no mocks, no network dependency — and wires it into CI alongside a new smoke test for `build/setupcli/jf.sh`, which previously had no coverage at all.

---

---

### DIFF
--- a/build/setupcli/jf.sh
+++ b/build/setupcli/jf.sh
@@ -63,10 +63,58 @@ else
     esac
 fi
 
-URL="https://releases.jfrog.io/artifactory/jfrog-cli/${CLI_MAJOR_VER}/${VERSION}/jfrog-cli-${CLI_OS}-${ARCH}/${FILE_NAME}"
+BASE_URL="${JFROG_CLI_RELEASES_BASE_URL:-https://releases.jfrog.io/artifactory/jfrog-cli}"
+URL="${BASE_URL}/${CLI_MAJOR_VER}/${VERSION}/jfrog-cli-${CLI_OS}-${ARCH}/${FILE_NAME}"
 echo "Downloading from: $URL"
-curl -XGET "$URL" -L -k -g > $FILE_NAME
-chmod +x $FILE_NAME
+curl -XGET "$URL" -L -g -o "$FILE_NAME"
+
+# Verify the download against the SHA256 checksum sidecar published for this
+# binary (JGC-542) before it is ever chmod +x'd or moved onto PATH. The
+# checksum lives at the exact same path as the binary, with .sha256 appended
+# (a per-binary sidecar, not a shared manifest - see JGC-542's corrected design).
+CHECKSUM_URL="${URL}.sha256"
+CHECKSUM_TMP="${FILE_NAME}.sha256.tmp"
+
+echo "Verifying checksum against: $CHECKSUM_URL"
+if ! curl -sS --fail -L -g "$CHECKSUM_URL" -o "$CHECKSUM_TMP"; then
+    echo "ERROR: could not download the checksum from $CHECKSUM_URL" >&2
+    echo "  (downloaded binary: $URL)" >&2
+    echo "Refusing to install an unverified $FILE_NAME binary." >&2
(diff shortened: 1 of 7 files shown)


build/setupcli/jf.sh
```diff
@@ -63,10 +63,58 @@ else
     esac
 fi
 
-URL="https://releases.jfrog.io/artifactory/jfrog-cli/${CLI_MAJOR_VER}/${VERSION}/jfrog-cli-${CLI_OS}-${ARCH}/${FILE_NAME}"
+BASE_URL="${JFROG_CLI_RELEASES_BASE_URL:-https://releases.jfrog.io/artifactory/jfrog-cli}"
```
vitaliil-jfrog: Are end users supposed to be able to override the URL? There is no protection from using a wrong url if `JFROG_CLI_RELEASES_BASE_URL` is set.

build/setupcli/jf.sh
```diff
@@ -63,10 +63,58 @@ else
...
 echo "Downloading from: $URL"
-curl -XGET "$URL" -L -k -g > $FILE_NAME
-chmod +x $FILE_NAME
+curl -XGET "$URL" -L -g -o "$FILE_NAME"
+
+# Verify the download against the SHA256 checksum sidecar published for this
+# binary (JGC-542) before it is ever chmod +x'd or moved onto PATH. The
+# checksum lives at the exact same path as the binary, with .sha256 appended
+# (a per-binary sidecar, not a shared manifest - see JGC-542's corrected design).
+CHECKSUM_URL="${URL}.sha256"
```
vitaliil-jfrog, line 75: Is this checksum available for older versions of the CLI? If not, users might not be able to download older versions.

build/setupcli/jf.sh
ehl-jf: Good catch — removed it. In practice it wasn't adding much risk beyond what's already there: this is a plaintext shell script someone downloads and runs directly, so nothing stops them (or an attacker with that level of access) from editing it before running it anyway — it was never a blackbox binary. But there's no reason to ship an unused override hook either, so `BASE_URL` is hardcoded now with no env var at all.

build/setupcli/jf.sh
ehl-jf: Tested this against a few old versions (2.50.0, 2.75.0, 2.122.0, 2.123.0) — checksums exist and match for all of them, e.g.:
```
curl https://releases.jfrog.io/artifactory/jfrog-cli/v2-jf/2.50.0/jfrog-cli-mac-arm64/jf.sha256
4c14906e00164a0592ec0e8506800f0b026147a5facc318ec875b7002d6a4897
```
matches `shasum -a 256` of the real 2.50.0 binary exactly. Seems like Artifactory serves this off checksum metadata it computes for every artifact at ingest, not off something JGC-542 had to backfill — so it works retroactively without us doing anything. Only version I found that 404s is 2.0.0, and that's because the binary itself was never published for that platform, not a checksum gap. Didn't add a version-based skip since there doesn't seem to be an actual gap to work around.

vitaliil-jfrog (approved):

PR #3717, opened 2026-09-21
RTECO-2247: Add jf Install-/Save-/Update-/Publish-PSResource commands

bhanurp:
## Summary

Part of RTECO-2247 — registers four top-level commands wrapping PowerShell's PSResourceGet module: `jf Install-PSResource`, `jf Save-PSResource`, `jf Update-PSResource`, `jf Publish-PSResource`. Deliberately four separate commands (matching native PowerShell cmdlet names) rather than one `jf psresource <verb>` wrapper — unlike `jf choco`, which uses a single-command, first-positional-argument dispatch.

Depends on (must merge first, or this repo's `go.mod` needs re-pinning to released versions once they land):
- build-info-go: https://github.com/jfrog/build-info-go/pull/430
- jfrog-cli-core: https://github.com/jfrog/jfrog-cli-core/pull/1620
- jfrog-cli-artifactory: https://github.com/jfrog/jfrog-cli-artifactory/pull/563

**Note on `go.mod`**: this branch carries `replace` directives pinning all three dependencies above to their own `RTECO-2247`(-psresource) branches (same canonical repos, not a fork) so this repo can build against the unreleased changes. Marked with an explicit release-blocking comment in `go.mod` — these need to come out once the sibling PRs merge and real released versions can be pinned.

## What's here

- `buildtools/cli.go` — the four `cli.Command` entries, built from a shared table (`psResourceCommandEntries`) rather than four hand-duplicated literals, all backed by one `psResourceCmd(cmdletName)` helper.
- `docs/buildtools/psresource/help.go` — per-cmdlet help text, rendered from one shared template parameterized by a small `cmdletMeta` struct (the four cmdlets' help is ~90% identical prose).
- `utils/cliutils/commandsflags.go` — the shared flag set (`--build-name`, `--build-number`, `--module`, `--project`, `--repo`, `--repo-resolve`, `--server-id`).
- Integration tests (`psresource_test.go`) — help text on every platform, CLI-level build-flag-pair validation, and (skipping gracefully without a local `pwsh` + PSResourceGet install) end-to-end dependency/artifact build-info collection for all four commands.
- CI workflow (`psresourceTests.yml`) — cross-platform (Linux/macOS/Windows), unlike `jf choco`'s Windows-only requirement, since PSResourceGet via `pwsh` is cross-platform.

## Notable bug fix included (found via an adversarial multi-agent code review after the initial implementation)

- **Curation audit silently disabled**: `WrapCmdWithCurationPostFailureRun` was called with each cmdlet's own PascalCase name (`"Install-PSResource"`, etc.) as `cmdName`, but jfrog-cli-security's post-failure curation audit gates on a fixed, generic verb allowlist (`{install, build, i, add, ci, get, mod}`) shared across every package manager — none of our names was ever in it, so the audit was a silent no-op for all four commands. Install/Save/Update now pass the matching `"install"` verb; Publish (which uploads rather than resolves a package, so curation can't block it the way this audit checks for) runs directly without a curation `cmdName` that would never apply.

## Test plan

- `go test . -run PSResource -v -args -test.psresource=true` and `go test ./buildtools/... ./docs/buildtools/psresource/...` — all pass (end-to-end tests skip gracefully on this machine, which has no `pwsh` installed).
- New regression tests for the curation-cmdName fix, `GetAIDescription`'s per-cmdlet map lookup, and dispatch-reaching coverage for Save/Update/Publish (previously only Install-PSResource's real dispatch logic was ever exercised past the shared `--help` early-return).
- `golangci-lint` clean, `gofmt` clean.

- [x] All tests passed. New tests added for the bug found during review.
- [x] All static analysis checks passed.
- [x] This pull request is on the master branch.
- [x] Full flow was locally tested (help text on every platform; end-to-end paths verified with `pwsh` unavailable, which is the graceful-skip path).

---
**Marked as draft**: this repo's required `No-Replace` check hard-fails while any `replace github.com/jfrog/*` directive is active in `go.mod` (regardless of target) — and this branch needs one for each of build-info-go#430, jfrog-cli-core#1620, and jfrog-cli-artifactory#563 until they merge and this branch can pin real released versions. Will mark ready for review once those merge and the replace directives come out.


.github/workflows/psresourceTests.yml
```diff
@@ -0,0 +1,66 @@
...
+        shell: bash
+        run: |
+          if ! command -v pwsh >/dev/null 2>&1; then
+            if [ "${{ matrix.os.name }}" = "ubuntu" ]; then
+              curl -sSL https://aka.ms/install-powershell.sh | sudo bash
+            else
+              brew install --cask powershell
+            fi
+          fi
+          pwsh -v
```
naveenku-jfrog: This can be avoided at all. PowerShell 7 (pwsh) is already pre-installed out of the box on all operating systems(standard GitHub-hosted runners (ubuntu-latest, macos-latest, or windows-latest)
```
- name: Verify PowerShell is installed
  run: pwsh -v
```

naveenku-jfrog (changes requested):

.github/workflows/psresourceTests.yml
bhanurp: Good catch, thanks! You're right — the matrix runs on standard GitHub-hosted runners (ubuntu-24.04, windows-2022, macos-14), which already ship PowerShell 7 as `pwsh` out of the box, so the install-if-missing logic (curl/brew) was unnecessary.

Simplified the step in 96c4148a to just:
```yaml
- name: Verify PowerShell is installed
  run: pwsh -v
```

Pushed to the branch.

naveenku-jfrog (approved):

